PR #39, opened 2016-10-26
Enable Braintree credit card payments via hosted fields

luukveenis:
This adds the ability to make credit card payments using Braintree's hosted fields. We provide both frontend and backend views, but at the moment sources are reusable only in the backend.


luukveenis:
I am :+1: on the changes @adammathys added, but would be good to get another set of eyes on this.

PR #44, opened 2016-11-20
Add Braintree configuration to Spree::Store model

luukveenis:
Since we're reusing the same payment method and source for Paypal, Apple Pay, and Braintree Hosted Fields, we need some way of configuring which payment methods are enabled on a per-store basis.

This adds a new `Configuration` model that's associated with stores as `braintree_configuration`, where we can set any necessary configuration options. I also added some simple views for admins to modify the configurations.

braintreeconfig

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,23 @@ bundle
 bundle exec rails g solidus_paypal_braintree:install
 ```
 
+Store Configuration
+-------------------
+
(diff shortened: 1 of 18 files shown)


luukveenis:
jenkins test this please

luukveenis:
jenkins test this please

luukveenis:
jenkins test this please

PR #45, opened 2016-11-21
Fix Apple Pay feature spec

luukveenis:
This test was passing under Selenium but failing about 70% of the time under poltergeist. The main reason is that the "Log In" button doesn't always register clicks immediately, so we need to retry until the JS that controls it is initialized.


bbuchalter:
This PR is named "Fix Apple Pay feature spec", but the name of the spec being fixed is `paypal_checkout_spec.rb`. What's the thinking here?

luukveenis:
Ah, sorry @bbuchalter. I've been going back and forth between PayPal / Apple Pay stuff too much the last few days. The spec being fixed is definitely the PayPal one.

PR #48, opened 2016-12-15
Pass order state to transaction import

luukveenis:
Users might not always want the order to be advanced all the way through to confirm. This allows a state parameter to be sent to the create action on the transactions controller to specify the desired end state for the order. If none is provided, the default value 'confirm' is used.

### DIFF
--- a/app/models/solidus_paypal_braintree/transaction_import.rb
+++ b/app/models/solidus_paypal_braintree/transaction_import.rb
@@ -34,7 +34,7 @@ module SolidusPaypalBraintree
       order.user
     end
 
-    def import!
+    def import!(end_state)
       if valid?
         order.email = user.try!(:email) || transaction.email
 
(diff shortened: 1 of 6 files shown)


luukveenis:
jenkins test this please

app/models/solidus_paypal_braintree/transaction_import.rb
```diff
@@ -69,10 +63,26 @@ def address
...
+
+      until order.state == end_state
+        order.next!
+        update_payment_total(payment) if order.payment?
+      end
+    end
+
+    def update_payment_total(payment)
+      unless payment.amount == order.total
+        payment.update!(amount: order.total)
```
adammathys: This is likely to cause some issues in any sort of split-payment setup. (e.g.: Store credits.)

PR #68, opened 2017-03-16
Feature/getting-braintree-data

jacquesporveau:
Pivotal: https://www.pivotaltracker.com/story/show/141433021

Braintree has a lot of information that could benefit the solidus admin and wouldn't it be nice to see it without leaving solidus? This PR is meant to be the first step giving us a helper to easily retrieve data from Braintree.


jacquesporveau:
Not quite sure what to say about this one. My tests are all passing locally. I've closed and opened the PR to run Travis several and have had different results each build from which Solidus version is failing with my PR.

PR #73, opened 2017-04-26
Spec fixes

Senjai:
The suite was fairly unreliable for me. Included details in the commit message, but it's one of the rare areas I think a sleep is necessary until we tackle the "disable the submit button on load" issue.


Senjai:
@tvdeyen Not quite good to go yet, still pretty flakey

PR #78, opened 2017-05-02
Coverage

Senjai:


spec/models/solidus_paypal_braintree/source_spec.rb
```diff
@@ -138,4 +138,70 @@
...
+  describe "#credit_card?" do
+    subject { described_class.new(payment_type: type).credit_card? }
+
+    context "when the payment type is CreditCard" do
+      let(:type) { "CreditCard" }
+
+      it { is_expected.to be true }
+    end
+
+    context "when the payment type is not PayPal" do
```
luukveenis: This should be `is not CreditCard`

spec/models/solidus_paypal_braintree/source_spec.rb
```diff
@@ -116,7 +116,7 @@
       it { is_expected.to be true }
     end
 
-    context "when the payment type is not PayPal" do
+    context "when the payment type is not Apply Pay" do
```
luukveenis: :thinking:

PR #79, opened 2017-05-03
Set source payment_type in cc hosted fields

tvdeyen:
The hosted fields does not set the payment_type value on credit card
sources. This leads to sources which type can’t be recognized anymore.

### DIFF
--- a/app/models/solidus_paypal_braintree/source.rb
+++ b/app/models/solidus_paypal_braintree/source.rb
@@ -10,6 +10,8 @@ module SolidusPaypalBraintree
 
     belongs_to :customer, class_name: "SolidusPaypalBraintree::Customer"
 
+    validates :payment_type, inclusion: [PAYPAL, APPLE_PAY, CREDIT_CARD]
+
     scope(:with_payment_profile, -> { joins(:customer) })
     scope(:credit_card, -> { where(payment_type: CREDIT_CARD) })
 
--- a/db/migrate/20170508085402_add_not_null_constraint_to_sources_payment_type.rb
+++ b/db/migrate/20170508085402_add_not_null_constraint_to_sources_payment_type.rb
@@ -1,0 +1,11 @@
+class AddNotNullConstraintToSourcesPaymentType < ActiveRecord::Migration
+  def change
+    reversible do |dir|
+      dir.up do
+        SolidusPaypalBraintree::Source.where(payment_type: nil).
+          update_all(payment_type: 'CreditCard')
+      end
+    end
+    change_column_null(:solidus_paypal_braintree_sources, :payment_type, false)
+  end
+end
--- a/lib/solidus_paypal_braintree/engine.rb
+++ b/lib/solidus_paypal_braintree/engine.rb
@@ -10,7 +10,7 @@ module SolidusPaypalBraintree
 
     initializer "register_solidus_paypal_braintree_gateway", after: "spree.register.payment_methods" do |app|
       app.config.spree.payment_methods << SolidusPaypalBraintree::Gateway
-      Spree::PermittedAttributes.source_attributes << :nonce
+      Spree::PermittedAttributes.source_attributes.concat [:nonce, :payment_type]
     end
 
     def self.activate
--- a/app/views/spree/shared/_braintree_hosted_fields.html.erb
+++ b/app/views/spree/shared/_braintree_hosted_fields.html.erb
@@ -21,5 +21,6 @@
   </div>
 
   <div class="clear"></div>
+  <input type="hidden" name="<%= prefix %>[payment_type]" value="<%= SolidusPaypalBraintree::Source::CREDIT_CARD %>">
   <input type="hidden" id="payment_method_nonce" name="<%= prefix %>[nonce]">
 </div>
--- a/spec/controllers/solidus_paypal_braintree/checkouts_controller_spec.rb
+++ b/spec/controllers/solidus_paypal_braintree/checkouts_controller_spec.rb
@@ -18,7 +18,7 @@ RSpec.describe SolidusPaypalBraintree::CheckoutsController, type: :controller do
               "payment_method_id" => payment_method.id,
               "source_attributes" => {
                 "nonce" => "fake-paypal-future-nonce",
-                "payment_type" => "PayPal"
+                "payment_type" => SolidusPaypalBraintree::Source::PAYPAL
               }
             }
           ],
--- a/spec/controllers/solidus_paypal_braintree/transactions_controller_spec.rb
+++ b/spec/controllers/solidus_paypal_braintree/transactions_controller_spec.rb
@@ -36,7 +36,7 @@ RSpec.describe SolidusPaypalBraintree::TransactionsController, type: :controller
       {
         transaction: {
           nonce: "fake-valid-nonce",
-          payment_type: "MonopolyMoney",
+          payment_type: SolidusPaypalBraintree::Source::PAYPAL,
           phone: "1112223333",
           email: "batman@example.com",
           address_attributes: {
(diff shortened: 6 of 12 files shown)


tvdeyen:
Dang. The Travis push hook gets triggered because I accidentally pushed my branch to this repo instead of my fork. I deleted the branch while the build was running. This is why I failed. But the PR hook triggered the correct build of my Fork and this passes 😅

tvdeyen:
Closing and reopen, to get correct Travis result

tvdeyen:
It turns out that Travis seem to have mayor caching issues. Sorry about that. Please ignore the `continuous-integration/travis-ci/push` result.

spec/features/braintree_credit_card_checkout_spec.rb
```diff
@@ -48,14 +48,20 @@
...
       end
       click_button("Place Order")
+    end
+
+    it "checks out successfully" do
       expect(page).to have_content("Your order has been processed successfully")
     end
+
+    it "sets the payment type of source to CreditCard" do
+      expect(SolidusPaypalBraintree::Source.last.payment_type).to eq('CreditCard')
```
Senjai: This assertion does not belong in feature spec. Could we do this at the unit test level? We should probably also avoid using last unless we're sure the system is only creating the one record.

lib/solidus_paypal_braintree/engine.rb
```diff
@@ -10,7 +10,7 @@ class Engine < Rails::Engine
 
     initializer "register_solidus_paypal_braintree_gateway", after: "spree.register.payment_methods" do |app|
       app.config.spree.payment_methods << SolidusPaypalBraintree::Gateway
-      Spree::PermittedAttributes.source_attributes << :nonce
+      Spree::PermittedAttributes.source_attributes.concat [:nonce, :payment_type]
```
Senjai: This value isn't whitelisted and can be modified by users. Further if we know the payment type in advance, we should probably avoid requiring it to be set in the form.

tvdeyen:
> To my knowledge, the "card_type" and friends should be delegated to Braintree, via the delegate call there. As for determining what type of payment the source represents, the source already includes several query methods for that, are those not working in this scenario?

The `payment_type` is a [custom database column](https://github.com/solidusio/solidus_paypal_braintree/blob/master/db/migrate/20160830061749_create_solidus_paypal_braintree_sources.rb#L6) from our side, so Braintree does not know about it. 

For PayPal and ApplePay sources we [set these value explicitly](https://github.com/solidusio/solidus_paypal_braintree/blob/master/app/assets/javascripts/spree/frontend/paypal_button.js#L117), but not for credit card payments. My pending spec e4e458914ee0ff6778982daabcb633f29ab6c57c shows that this is currently not working. 

This is why I chose to test this integrative in stead of on the unit level. The model is polymorphic (not in the AR kind of things, but I plays multiple roles). We can't know the value upfront, so testing it on the unit level did not really made sense for me, we need to set this value while creating it through the controller. Therefore passing it along the params from the view - where we exactly know what type the source is - made the most sense for me.

Any idea how to approach this differently?

spec/features/braintree_credit_card_checkout_spec.rb
tvdeyen: > We should probably also avoid using last unless we're sure the system is only creating the one record.

We know exactly that in this scenario the last source is the one we created through the controller. I had a version where I explicitly tested that we only have on source here, but this seemed overkill for me. I can add this if you want to be extra save.

lib/solidus_paypal_braintree/engine.rb
tvdeyen: > This value isn't whitelisted and can be modified by users. 

I need to whitelist it, otherwise I can not pass it into the controller from the view where the value needs top be set, as we don't know the type upfront.

> Further if we know the payment type in advance, we should probably avoid requiring it to be set in the form.

We don't know the type upfront. I didn't find any code that guarantees this. Maybe you can point me to?

tvdeyen:
> the source already includes several query methods for that

The query methods [depend on the `payment_type` to be set](https://github.com/solidusio/solidus_paypal_braintree/blob/master/app/models/solidus_paypal_braintree/source.rb#L44-L54). `friendly_payment_type` even blows up if no `payment_type` is set.

tvdeyen:
The "best solution" would be to actually provide three payment sources (and even gateways). Then we wouldn't even need to have these query methods at all. 

I will provide a RFC for this, as I think now that we have a common interface for payment sources in Solidus 2.2 we should start to embrace it. For older stores its an easy back port we would ship with this gem.

lib/solidus_paypal_braintree/engine.rb
Senjai: >I need to whitelist it, otherwise I can not pass it into the controller from the view where the value needs top be set, as we don't know the type upfront.

I meant that the value isn't whitelisted. I can insert "lol" into this field and it would save correctly. We should probably add an inclusion validation to the model to prevent this.

>We don't know the type upfront. I didn't find any code that guarantees this. Maybe you can point me to?

Once we have the token, `payment_method.braintree.payment_method.find(source.token).is_a?(Braintree::PaypalAccount)`

As we only use this for display purposes, I no longer think that's worth doing though.

spec/features/braintree_credit_card_checkout_spec.rb
Senjai: Feature tests should only ever test user interactions. Since there is no reliable way to ensure that the user can see that this is the case, it should not be included in a feature test. If there was a way for a user to use the site in a manner that would cause a failure, we could test that here. Otherwise this should be tested in a controller/request test.

Senjai:
RE: the query methods, yes you're correct. Missed that, I should have looked at that more closely. I'm :+1: on including on the form, would just prefer the test moved.

As for the RFC can you hold off? We've had a couple discussions on that architecture and have the beginnings of a plan in place. If you're interested we can chat about it, but I don't want to duplicate efforts here.

tvdeyen:
Build fixes waits for #84

db/migrate/20170508085402_add_not_null_constraint_to_sources_payment_type.rb
```diff
@@ -0,0 +1,11 @@
+class AddNotNullConstraintToSourcesPaymentType < ActiveRecord::Migration
+  def change
+    reversible do |dir|
+      dir.up do
+        SolidusPaypalBraintree::Source.where(payment_type: nil).
+          update_all(payment_type: 'CreditCard')
```
Senjai: This isn't accurate. Is there a scenario that generated nil payment_types in the past?

app/models/solidus_paypal_braintree/source.rb
```diff
@@ -10,6 +10,8 @@ class Source < ApplicationRecord
 
     belongs_to :customer, class_name: "SolidusPaypalBraintree::Customer"
 
+    validates :payment_type, presence: true
```
Senjai: `validates :payment_type, inclusion: [PAYPAL, APPLE_PAY, CREDIT_CARD]`

db/migrate/20170508085402_add_not_null_constraint_to_sources_payment_type.rb
tvdeyen: Yes. All credit card payments. PayPal and ApplePay payments have this value set via JS tailored params. 

Although this isn't accurate it's the only thing we can do about, besides not doing this at all of course.

Senjai: Sure, fair point.

app/models/solidus_paypal_braintree/source.rb
tvdeyen: Very good point! 👌🏻

Senjai:
@tvdeyen Have some vcr issues no idea why though :/

tvdeyen:
@Senjai yeah. Pretty strange. Only for Solidus < 2.1. Even locally. Investigated this morning, but gave up because lack of time.

Senjai:
Closed in favor of #91 thanks thomas.

PR #80, opened 2017-05-04
Set source payment_type in cc hosted fields

tvdeyen:
The hosted fields does not set the payment_type value on credit card
sources. This leads to sources which type can’t be recognized anymore.


tvdeyen:
Ok, I'm out of ideas. Travis got confused by my accidentally wrong push to this repo instead to my fork.

PR #84, opened 2017-05-08
Always use the same zipcode in address factory

tvdeyen:
The Solidus address factory randomizes the zipcode. The OrderWalkThrough we use in the credit card checkout spec uses this factory for the user addresses.

For credit card payments we transmit the billing address to braintree, for paypal payments the shipping address.

As we match the body in our VCR settings VCR can not match the request anymore and therefore cannot replay existing cassettes. This leads to sporadic failing tests.

By modifying the spree address factory we ensure that the zipcode is always the same.

### DIFF
--- a/lib/solidus_paypal_braintree/factories.rb
+++ b/lib/solidus_paypal_braintree/factories.rb
@@ -4,3 +4,14 @@ FactoryGirl.define do
   # Example adding this to your spec_helper will load these Factories for use:
   # require 'solidus_paypal_braintree/factories'
 end
+
+FactoryGirl.modify do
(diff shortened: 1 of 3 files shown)


tvdeyen:
😞

Senjai:
We'll need a comment explaining why this change needs to be there.

tvdeyen:
@Senjai added a comment. 

> Could we also change this in core?

👍 I also thought this might be a good idea. Will open a PR.

tvdeyen:
> Will open a PR.

It turns out that we want to have the address factory support zipcodes for various countries.

PR #85, opened 2017-05-11
Scoped store

Senjai:
Gateways should be looked up only for the current store on the frontend.


tvdeyen:
Hence the STI nature of payment methods in Solidus, payment method ids are unique across stores. I doubt we need this change. 

Could we close this for now?

tvdeyen:
Closing as stale.

PR #86, opened 2017-05-12
Use flashes for displaying frontend errors

Senjai:
Allows users to override by replacing the javascript function before
rendering the page.

This is not great, and is being replaced on my js refactor branch, but
it's required for projects now.


Senjai:
Executive merging this. Bundler having their weekly infinite gem resolution issues.

PR #87, opened 2017-05-12
Attempt to stop infinite dep resolution

Senjai:
For some reason bundlers depedency graph has been terribly slow to
resolve. 20 minutes slow. Maybe this is fixed with the pre release, I
have no idea.

Bundler keeps breaking all the things sans changes to the project.


jhawthorn:
👍 This is the reason we need the rails `~> 4.2` in there anyways 😞

PR #88, opened 2017-05-15
Asset pipeline fixes

tvdeyen:
Without having an extension Rails refuses to precompile the assets
and raises an error in development mode.

### DIFF
--- a/lib/solidus_paypal_braintree/engine.rb
+++ b/lib/solidus_paypal_braintree/engine.rb
@@ -19,8 +19,6 @@ module SolidusPaypalBraintree
       end
     end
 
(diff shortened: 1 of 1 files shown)


Senjai:
@tvdeyen This appears to work for me though. If the assets weren't compiled, the test app shouldn't be able to find them. Is this just for production or?

tvdeyen:
> raises an error in development mode.

The app I'm using this in is a Rails 4.2 app. But this should also be an issue with Rails 5. Not shure why this isn't happening in the test app. Maybe Rails silently ignores this in test mode? Anyhow, it works with the extension and fixed the error in the app I'm developing.

PR #94, opened 2017-06-15
Support Rails 5.1

Sinetheta:


tvdeyen:
Bundler 😞

Sinetheta:
Heh, I don't think it's all bundler. I had local failures as well that look like they are due to changes to rake.

Issue #74, opened 2017-04-28
Payments in backend not working


tvdeyen:
Payments are not working in backend currently:

1. The JS only inits if previous cards are present
2. The Braintree client JS is not loaded/included in the backend

Got a fix in my client app, will provide a PR soon.

tvdeyen:
Here is my current branch: https://github.com/tvdeyen/solidus_paypal_braintree/tree/backend-payments

Lacks of tests, though.

isaacfreeman:
@tvdeyen: I've cherry-picked from your branch and it's working for me here so far. If it's OK with you, I'll continue with adding tests and get a PR up.

tvdeyen:
@isaacfreeman sure, thanks.

isaacfreeman:
@tvdeyen: I think we should be able to close this now.

tvdeyen:
Aight

PR #95, opened 2017-06-18
Support Rails 5.1

isaacfreeman:
Reworking #94 

Currently down to three build errors.

PR #97, opened 2017-06-21
Backend payments

isaacfreeman:
Per #74, set up backend payments. Practically all @tvdeyen's work so far.


tvdeyen:
@isaacfreeman I think we need a rebase after #95 was merged

isaacfreeman:
Progress: building green for master and v2.2.

lib/views/backend/spree/admin/payments/source_forms/_paypal_braintree.html.erb
```diff
@@ -1,10 +1,12 @@
 <% id = payment_method.id %>
+<% radio_buttons_use_name = Gem::Version.new(Spree.solidus_version) < Gem::Version.new('1.3') %>
+<% radio_button_id = radio_buttons_use_name ? 'braintree' : payment_method.id %>
```
tvdeyen: Since these views are only put into Rails’ loadpath, if `backend` is available, we could make two versions of this file and decide in `engine.rb` which one to pick.

isaacfreeman: Makes sense to me. If we have to check versions, better to do it once.

lib/views/backend_v1.2/spree/admin/payments/source_forms/_paypal_braintree.html.erb
```diff
@@ -0,0 +1,16 @@
+<% id = payment_method.id %>
+
+<fieldset class="no-border-bottom braintree-hosted-fields" data-braintree-hosted-fields data-id="braintree" data-payment-method-id="<%= id %>">
```
Sinetheta: Is `data-id="<%= id %>"` vs `data-id="braintree"` the only difference between these files? Can we move that out into ruby land where it's easier to understand and share the same partial?

isaacfreeman: Sure. It's a significant difference for Solidus version compatibility, but all the more reason to give it a named method.

isaacfreeman: On discussion, we agreed to leave this one as is. We anticipate that support for Solidus v1.2 may be dropped before too long, at which point it goes away.

tvdeyen: @Sinetheta like discussed earlier we don't want to have Solidus version comparison in view files. This file will be gone soon as we drop support for Solidus < 1.2

We could move all the parts that we share into a partial and only leave the wrapper in these two partials, but again, this is just an intermediate solution.

tvdeyen:
Thanks for picking up the work @isaacfreeman

PR #100, opened 2017-06-27
Setup recipe in README.md

isaacfreeman:
Rearranges README.md to provide a more straightforward setup recipe.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,46 +21,54 @@ bundle
 bundle exec rails g solidus_paypal_braintree:install
 ```
 
-Usage
------
+## Basic Setup
 
-This gem extends Solidus by providing a new payment method and source, named
-`SolidusPaypalBraintree::Gateway` and `SolidusPaypalBraintree::Source` respectively.
-All payment types - PayPal, ApplePay, and Credit Cards - are supported through
-the same payment method.
+### Retrieve Braintree account details
+You'll need the following account details:
+- `Merchant ID`
+- `Public key`
+- `Private key`
 
-The payment method requires 3 preferences to be set to process payments:
-- `merchant_id`
-- `public_key`
-- `private_key`
-
-These values can be obtained by logging in to your Braintree account and going
+These values can be obtained by logging in to your Braintree account, going
 to `Account -> My User` and clicking `View Authorizations` in the **API Keys,
 Tokenization Keys, Encryption Keys** section.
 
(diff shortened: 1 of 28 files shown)


README.md
```diff
@@ -21,46 +21,54 @@ bundle
...
+```ruby
+SolidusPaypalBraintree::Gateway.new(
+  name: "Braintree",
+  preferences: {
+    public_key:  '[your public key]',
+    private_key: '[your private key]',
+    merchant_id: '[your merchant ID]'
+  }
+).save
+```
```
tvdeyen: I would like us to promote static preferences over database stored credentials.

```ruby
# config/initializers/spree.rb
Spree::Config.configure do |config|
  config.static_model_preferences.add(
    SolidusPaypalBraintree::Gateway,
    'braintree_credentials', {
      environment: Rails.env.production? ? 'production' : 'sandbox',
      merchant_id: ENV['BRAINTREE_MERCHANT_ID'],
      public_key: ENV['BRAINTREE_PUBLIC_KEY'],
      private_key: ENV['BRAINTREE_PRIVATE_KEY']
    }
  )
end
```

README.md
```diff
@@ -21,46 +21,54 @@ bundle
...
-This preference allows users to provide different Merchant Account Ids for
-different currencies. If you only plan to accept payment in one currency, the
-defaut Merchant Account Id will be used and you can omit this option.
-An example of setting this preference can be found
-[here](https://github.com/solidusio/solidus_paypal_braintree/blob/master/spec/spec_helper.rb#L70-L72).
+### Create a new payment method
+1. Visit `/admin/payment_methods/new`.
+2. Set `provider` to SolidusPaypalBraintree::Gateway
+3. Enter the three required preferences: `merchant_id`, `public_key`, `private_key`. Other optional preferences are discussed below.
+4. Click `Update` to save
```
tvdeyen: If we want to use static preferences (what we should IMO - see below) we need to change this to:

3. Click "Save"
4. Choose `braintree_credentials` from "Preference Source" select
5. Hit "Update" to save

README.md
isaacfreeman: OK, makes sense. I think we need to cover both approaches, since they're both provided. But I'll write it to indicate that static preferences are preferred.

isaacfreeman:
Merge after #105

tvdeyen:
Thanks

PR #90, opened 2017-05-16
[WIP] JS Overhaul

Senjai:
An uncomplete (with changes + comments pending from other branches) redo of the javascript with extensibility and namespaces in mind.

Feedback welcome now, but this PR is not final.

PR #101, opened 2017-06-29
JS Overhaul

isaacfreeman:
Continues work by @Senjai in #90


isaacfreeman:
Added recommendations from @tvdeyen's review of #90

tvdeyen:
Still 👍 

Would love to see this great work of @Senjai merged finally. Would love to see this tested with ApplePay, though. PayPal and Credit Cards are confirmed working for me, but I can't test Apple Pay right now.

isaacfreeman:
At this point all that remains is to test with Apple Pay. I'll be able to continue with that when my MacBook Pro is repaired. ETA... maybe a week.

app/assets/javascripts/solidus_paypal_braintree/client.js
```diff
@@ -262,20 +263,20 @@ SolidusPaypalBraintree.Client.prototype.initializeApplePaySession = function(con
   session.begin();
 },
 
-SolidusPaypalBraintree.Client.prototype._buildTransaction = function(payload, config, shippingContact) {
+SolidusPaypalBraintree.Client.buildTransaction = function(payload, config, shippingContact) {
```
Senjai: Why were these brought out of the prototype? They shouldn't be static no?

app/assets/javascripts/solidus_paypal_braintree/client.js
```diff
@@ -0,0 +1,186 @@
...
+ *
+ * We use this class mostly to hide the token operations for users.
+ *
+ * After creating the class, a call should be made to initialize before using it.
+ * @see initialize
+ *
+ * @constructor
+ * @param {Object} config Initalization options for the client
+ * @param {Boolean} config.useDataCollector Use data collector capabilities for the braintree client
+ * @param {Boolean} config.usePapyal Use Paypal capabilities for the braintree client
```
Sinetheta: typo

app/assets/javascripts/solidus_paypal_braintree/client.js
isaacfreeman: It made it easier to reference them from within `session.onpaymentauthorized`. But yes, this was intended to be a temporary step while I refactored other code. On my list to fix up before I remove the '[WIP]' from the title.

isaacfreeman:
The last thing I want to do here is update README.md, but I'd like to get #100 merged first.

app/assets/stylesheets/spree/frontend/solidus_paypal_braintree.css
```diff
@@ -16,14 +16,24 @@ the installer will append this file to the app vendored assets here: 'vendor/ass
...
+  background-position: 50% 50%;
+  border-radius: 5px;
+  padding: 0;
+  margin: 5px auto;
+  transition: background-color .15s;
+  width: auto;
+}
+
+.apple-pay-button.visible {
+  visibility: visible;
```
tvdeyen: I think we should use `display` over `visibility` as the latter still occupies space, while the former not.
We don't want to have holes in the layout if the payment va Apple Pay is not possible. But this can also be tackled in a later PR and should not hold this already very large PR back.

tvdeyen:
Before merging this we should think about cutting a release first. As this will most likely brake lots of stores that have customized the frontend JS (like I have).

app/assets/stylesheets/spree/frontend/solidus_paypal_braintree.css
isaacfreeman: This is why I favour `visibility`! At least on my machine, it takes a noticeable amount of time for the Apple Pay button to appear. So when it does, it makes everything below jump down, which doesn't draw the eye to the button itself.

tvdeyen: Interesting as the image is bundled inside the browser itself. Lots of users (the vast majority) will never see the button ever as they don't have an Apple browser, Phone or live in a country where Apple Pay is not available. So the mayor use case of this button is actually do not be seen ever. We should display a "You're browser does not support Apple Pay" notice instead.

This is the approach I went with:

```css
#apple-pay-button {
  background-image: -webkit-named-image(apple-pay-logo-white);
  background-position: center;
  background-size: 60%;
  background-repeat: no-repeat;
  border-radius: 5px;
}

#apple-pay-button.supported {
  display: inline-block;
  width: 160px;
  height: 40px;
  font-size: 0;
  background-color: black;
  cursor: pointer;
}
```

```html
<span id="apple-pay-button">
  Your browser does not support ApplePay.
</span>
```

```js
...
braintree.fetchToken(function(clientToken) {
  braintree.initialize(clientToken, function(braintreeClient) {
    braintree.setupApplePay(braintreeClient, "<%= ENV['APPLE_PAY_MERCHANT_ID'] %>", function(applePayInstance) {
      $applePayButton.addClass('supported');
      ...
    });
  });
});
```

So, only for users that actually should see this button it appears, for all other people the notice appears.

isaacfreeman: I'd rather not have a message if the browser doesn't support Apple Pay. Most browsers aren't ever going to support it, so it seems like unnecessary information to put in front of a customer. It seems better to me that it just doesn't appear unless it's available to use.

In any case, I agree that any changes here should be in a separate PR.

isaacfreeman:
All green in Travis

Remaining to do on this:
- decide on versioning
- do another round of manual Apple Pay testing

app/assets/stylesheets/spree/frontend/solidus_paypal_braintree.css
tvdeyen: > I'd rather not have a message if the browser doesn't support Apple Pay

Also fine. But it should not occupy space then, so I still vote for `display` over `visibility` in a separate PR

PR #107, opened 2017-07-12
Show transaction link

isaacfreeman:
When we click through to a payment in admin, include a link to the Braintree admin page for the payment. 

This is based on https://github.com/solidusio/solidus_braintree/commit/68abb9683bf38a546ae703df7433a4e8233b9c81 by @ericsaupe

### DIFF
--- a/config/initializers/braintree.rb
+++ b/config/initializers/braintree.rb
@@ -1,0 +1,1 @@
+Spree::Admin::PaymentsController.helper :braintree_admin
--- a/app/helpers/braintree_admin_helper.rb
+++ b/app/helpers/braintree_admin_helper.rb
@@ -1,0 +1,18 @@
+module BraintreeAdminHelper
+  # Returns a link to the Braintree web UI for the given Braintree payment
+  def braintree_transaction_link(payment)
+    environment = payment.payment_method.preferred_environment == 'sandbox' ? 'sandbox' : 'www'
+    merchant_id = payment.payment_method.preferred_merchant_id
+    response_code = payment.response_code
+
+    return unless response_code.present?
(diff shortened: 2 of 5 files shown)


config/initializers/braintree.rb
```diff
@@ -0,0 +1 @@
+ActionView::Base.send :include, BraintreeViewHelpers
```
tvdeyen: Could we include this in the admin payments controller instead?

```ruby
Spree::Admin::PaymentsController.helper :braintree_admin_helpers
```

app/helpers/braintree_view_helpers.rb
```diff
@@ -0,0 +1,18 @@
+module BraintreeViewHelpers
```
tvdeyen: Could we rename this into `BraintreeAdminHelpers` or something similar that indicates this helper is used for admin views?

app/views/spree/admin/payments/source_views/_paypal_braintree.html.erb
```diff
@@ -1,10 +1,15 @@
 <fieldset data-hook="credit_card">
+  <legend align="center"><%= t('braintree', scope: 'solidus_paypal_braintree') %></legend>
```
tvdeyen: Why this `legend`? No other admin payment view has this legend and admins do not have to know about the provider of this payment method IMO. But I'm not considering this a blocker.

isaacfreeman: I didn't want it to say "Credit Card", because it might be Apple Pay or Paypal instead. But you've prompted me to rewrite it with a better localisation call.

PR #110, opened 2017-07-23
Set Braintree log level as a gateway preference

isaacfreeman:
The `braintree` gem provides a default logger that logs at the INFO level, which clutters logs a bit.

This provides a gateway preference for log level, and sets the default a notch higher to WARN.


tvdeyen:
Why don't we use the current debug level of the application?

tvdeyen:
@isaacfreeman do you think it makes sense to use the log level of the rails application?

PR #111, opened 2017-07-23
Fix deprecated code

isaacfreeman:
Commits to address deprecation warnings that are showing up in specs.


tvdeyen:
Spec failures are random. Will rebuild until it passes

PR #113, opened 2017-07-27
Use the shared spec_helper from solidus_support

tvdeyen:
This should help us remove duplication and fix common spec problems
like asset preloading.


tvdeyen:
Github has problems picking up the force push

PR #114, opened 2017-07-27
Remove integer type declaration from cassettes

tvdeyen:
Lots of older cassettes had `type="integer"` attributes in the xml
of the request body. Newer cassettes do not have this. This causes
build errors depending on which machine the specs run.

It is not 100% clear what's the cause of this.

Maybe the builder gem? - braintree depends on >=2.0.0 while 3.2.3 is the most recent version
Maybe the braintree gem? - 2.76.0 is the most recent version and was updated lately
Maybe libxml? - The version on the different machines may vary

### DIFF
--- a/spec/fixtures/cassettes/admin/invalid_credit_card.yml
+++ b/spec/fixtures/cassettes/admin/invalid_credit_card.yml
@@ -8,7 +8,7 @@ http_interactions:
       string: |
         <?xml version="1.0" encoding="UTF-8"?>
         <client-token>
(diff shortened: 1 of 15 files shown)


tvdeyen:
It would be great to know the root cause of the type declaration being sometimes present sometimes not - dependent on where the cassettes have been generated.

tvdeyen:
The root cause is Ruby 2.4 not having a Bignum and Fixnum class anymore. https://github.com/braintree/braintree_ruby/pull/143

PR #115, opened 2017-07-28
Better error handling

tvdeyen:
This adds several improvements to error handling of transaction responses.

- Record AVS and CVV results
- Include request params in error responses
- Humanize transaction error messages

### DIFF
--- a/app/models/solidus_paypal_braintree/response.rb
+++ b/app/models/solidus_paypal_braintree/response.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 # Response object that all actions on the gateway should return
 module SolidusPaypalBraintree
   class Response < ActiveMerchant::Billing::Response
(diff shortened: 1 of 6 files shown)


config/locales/en.yml
```diff
@@ -28,3 +28,16 @@ en:
       tab: Braintree
       update_success: Successfully updated Braintree configurations.
       update_error: An error occurred while updating Braintree configurations.
+    gateway_rejection_reasons:
+      avs: AVS check failed.
+      avs_and_cvv: AVS and CVV check failed.
+      cvv: CVV check failed.
+      duplicate: Duplicated transaction.
```
mamhoff: Duplicate Transaction

app/models/solidus_paypal_braintree/response.rb
```diff
@@ -18,18 +19,14 @@ def build(result)
...
-        avs_result = nil
-        cvv_result = nil
-
         options = {
-          test: test,
-          authorization: authorization,
-          fraud_review: fraud_review,
-          avs_result: avs_result,
-          cvv_result: cvv_result
+          test: !Rails.env.production?,
```
mamhoff: Rather than `true` or `!Rails.env.production`, this should either be `nil` (indicating we have no idea) or something braintree tells us about using a sandbox.

tvdeyen: Unfortunately we don't have access to anything from either Braintree nor from Solidus to set this value correctly. ActiveMerchant sets this value to `false` per default. As nobody consuming this response actually uses this value I think we can just remove it and rely on the ActiveMerchant default behavior here.

PR #116, opened 2017-08-14
Respect Spree::Config[:address_requires_state]

isaacfreeman:


lib/views/frontend/spree/checkout/payment/_paypal_braintree.html.erb
```diff
@@ -37,7 +37,7 @@
       line1: '<%= address.address1 %>',
       line2: '<%= address.address2 %>',
       city: '<%= address.city %>',
-      state: '<%= address.state.name %>',
+      <% if Spree::Config[:address_requires_state] %> state: '<%= address.state.name %>',
```
tvdeyen: I guess we are missing an `<% end %>` here

tvdeyen:
@isaacfreeman still working on this? I think the fix is reasonable and worth to be merged, but we need to fix the syntax error. Also a spec would be great.

isaacfreeman:
Updated, but don't merge this yet – I'll add a spec, as recommended.

seand7565:
Hello from 3 years in the future! The file this PR is attempting to fix has since changed, and no longer carries the address. The address instead lives here: https://github.com/solidusio/solidus_paypal_braintree/blob/master/lib/views/frontend/spree/shared/_paypal_checkout_button.html.erb

#237 is more recent and also handles this issue - in a way that pulls the address logic out of the view, which I like. I think we can close this PR in favor of #237 .

PR #118, opened 2017-08-17
Update braintree client libs to v3.22.1

tvdeyen:
We use rather outdated (and even inconsistent) versions of Braintree JS libs.


tvdeyen:
This is a good idea. I actually felt in that trap today. Expect a PR soonish

PR #120, opened 2017-08-30
Don't try and add billing address if there is no billing address atta…

joeljackson:
…ched to credit card

### DIFF
--- a/app/models/solidus_paypal_braintree/gateway.rb
+++ b/app/models/solidus_paypal_braintree/gateway.rb
@@ -266,7 +266,7 @@ module SolidusPaypalBraintree
         params[:shipping] = braintree_shipping_address(options)
       end
 
-      if source.credit_card?
+      if source.credit_card? && options[:bliing_address].present?
         params[:billing] = braintree_billing_address(options)
       end
 


app/models/solidus_paypal_braintree/gateway.rb
```diff
@@ -266,7 +266,7 @@ def transaction_options(source, options, submit_for_settlement = false)
         params[:shipping] = braintree_shipping_address(options)
       end
 
-      if source.credit_card?
+      if source.credit_card? && options[:bliing_address].present?
```
tvdeyen: It would be great to have a test for this.

Then you actually would have found your typo as well ;)

tvdeyen:
@joeljackson I still like this change. Are you able to provide a spec and rebase with latest `master` to get rid of the merge conflicts and please remove the commits from #123 

Then we are fine. Thanks for your contribution.

joeljackson:
Hi @tvdeyen, still interested in this change?

If so would be interested in adding the spec and rebasing.

app/models/solidus_paypal_braintree/gateway.rb
```diff
@@ -306,11 +306,11 @@ def transaction_options(source, options, submit_for_settlement = false)
         params[:payment_method_nonce] = source.nonce
       end
 
-      if source.paypal?
+      if source.paypal? && options[:billing_address].present?
```
tvdeyen: Shouldn't this be 

```rb
options[:shipping_address].present?
```

lib/views/frontend/spree/checkout/existing_payment/_paypal_braintree.html.erb
```diff
@@ -0,0 +1,15 @@
...
+  <td>
+    <%= radio_button_tag "order[wallet_payment_source_id]", wallet_payment_source.id, default, class: "existing-cc-radio" %>
+  </td>
+  <td>
+    <%- if wallet_payment_source.payment_source.credit_card? %>
+      <div><%= wallet_payment_source.payment_source.cardholder_name %></div>
+      <div><%= wallet_payment_source.payment_source.card_type %> - <%= wallet_payment_source.payment_source.display_number %></div>
+      <div><%= wallet_payment_source.payment_source.expiration_month %>/<%= wallet_payment_source.payment_source.expiration_year %></div>
+    <%- elsif wallet_payment_source.payment_source.paypal? %>
+      <div>Paypal</div>
```
tvdeyen: Should be PayPal (Pal is also capitalized)

joeljackson: Oh, Shoot.  Didn't mean to include this here.

Will pull it out. See other pull request to see if we want to pull that in.

Thanks!

Issue #122, opened 2017-08-30
Allow sources to be added to the wallet on charge


joeljackson:
Looks like all that's needed is to implement a "reusable?" method on SolidusPaypalBraintree::Source
and we're good to go.

tvdeyen:
Fixed with merge of #123

PR #123, opened 2017-09-01
Allows the paypal_braintree source to be added to the wallet on checkout

joeljackson:


joeljackson:
Looking for some feedback here. This is intended to address https://github.com/solidusio/solidus_paypal_braintree/issues/122.

Wanted to ensure a source is always reusable and that making it a subclass of Spree::PaymentSource wasn't going to ruin anyone's day before proceeding.

swcraig:
Although it would make sense, I don't think that subclassing `Spree::PaymentSource` is the way to go here. `Spree::PaymentSource` [was only introduced in Solidus v2.2](https://github.com/solidusio/solidus/commit/c08f334e97cead1c10eb66fc1051e6d2a18b84f9), so subclassing it will [break this extension with all versions of Solidus before v2.2](https://travis-ci.org/solidusio/solidus_paypal_braintree/builds/270688825?utm_source=github_status&utm_medium=notification).

I believe your other two changes are what is needed to get this source added to the wallet.

tvdeyen:
Unfortunately it’s needed to subclass from PaymentSource because Solidus introduced a validation on the payment source class in https://github.com/solidusio/solidus/blob/master/core/app/models/spree/wallet_payment_source.rb. 

We either should remove this validation (as this is useless IMO) or need to change this code so it only subclasses on Solidus 2.2 and above.

joeljackson:
Seems like https://github.com/solidusio/solidus/pull/2201 should allow us to not subclass the PaymentSource. Will make appropriate changes.

joeljackson:
Notably thinking about this it needs to be a subclass of PaymentSource on Solidus 2.3 or this wont

app/models/solidus_paypal_braintree/source.rb
```diff
@@ -1,12 +1,13 @@
 module SolidusPaypalBraintree
-  class Source < ApplicationRecord
+  class Source < Spree::PaymentSource
```
jordan-brough: Can you change this to `SolidusSupport.payment_source_parent_class` please?
See https://github.com/solidusio/solidus_support/pull/1

app/models/solidus_paypal_braintree/source.rb
```diff
@@ -1,12 +1,13 @@
...
     PAYPAL = "PayPalAccount"
     APPLE_PAY = "ApplePayCard"
     CREDIT_CARD = "CreditCard"
 
     belongs_to :user, class_name: Spree::UserClassHandle.new
     belongs_to :payment_method, class_name: 'Spree::PaymentMethod'
     has_many :payments, as: :source, class_name: "Spree::Payment"
+    has_many :wallet_payment_sources, class_name: 'Spree::WalletPaymentSource', as: :payment_source, inverse_of: :payment_source
```
jordan-brough: You shouldn't need this if you inherit from `SolidusSupport.payment_source_parent_class`

jordan-brough:
@joeljackson can you please also update the gemspec to require at least version `0.1.2` of solidus_support?

joeljackson:
@jordan-brough Done.  Thanks 👍

jordan-brough:
Thanks @joeljackson. Lgtm. Seeing if we can figure out what's wrong w/ the spec failures.

PR #126, opened 2017-10-19
Protect against connection errors

tvdeyen:
Neither the `braintree` gem nor we handle various connection errors Rubys `net/http` library raises. Solidus expects [1] that payment method implementations raise an `ActiveMerchant::ConnectionError` if one of the several errors [2] happen and handles the payment accordingly. Without that protection we leave payments in processing state whenever a connection error happens.

This uses the activemerchant/network_connection_retries module to prtoect us against these errors and with that ensure that we always have a `ActiveMerchant::ConnectionError` that Solidus knows how to handle correctly.

[1] https://github.com/solidusio/solidus/blob/master/core/app/models/spree/payment/processing.rb#L198-L202
[2] https://github.com/activemerchant/active_merchant/blob/master/lib/active_merchant/network_connection_retries.rb#L4-L12

### DIFF
--- a/app/models/solidus_paypal_braintree/avs_result.rb
+++ b/app/models/solidus_paypal_braintree/avs_result.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'active_merchant/billing/avs_result'
+
 module SolidusPaypalBraintree
   class AVSResult < ActiveMerchant::Billing::AVSResult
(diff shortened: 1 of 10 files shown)


mamhoff:
We need to protect against more errors, also in the `BraintreeSource` class.

tvdeyen:
Extracted the PayPal spec hardening commit into #127

PR #127, opened 2017-10-19
Try to harden the PayPal integration spec

tvdeyen:
This spec fails every time on Travis because it interacts with the real PayPal popup
window. This can not be stable at all. Pending the spec when the capybara max wait time
exceeds in stead.


tvdeyen:
With the build errors we just had I am more and more in favor of removing this spec all together. This will never be a stable spec. We don’t control a huge part of it.

PR #129, opened 2017-10-23
Add more versions and test both DBs on travis

jhawthorn:
travis.yml was missing both v2.3 and v2.4


tvdeyen:
Dependencies added

PR #130, opened 2017-10-24
Pend paypal spec if paypal popup not available

tvdeyen:
It can happen that the paypal popup does not appear. We should not fail our spec when this happens.


tvdeyen:
Includes a fix for the hard coded ID in the client tokens controller spec as well.

PR #133, opened 2017-11-10
Support 18n for Braintree error messages

Gamabunta94:
With this PR I want to add the localization for Braintree error messages.
For this purpose I add the Braintree error messages in the localization rails file and show it with the corresponding slug returned from the Braintree response.
I also added Italian translation of the messages.


Gamabunta94:
It seems the test fails for a timeout error or something else unrelated.

app/assets/javascripts/solidus_paypal_braintree/constants.js
```diff
@@ -6,14 +6,11 @@ SolidusPaypalBraintree = {
       clientTokens: Spree.pathFor('solidus_paypal_braintree/client_token'),
       transactions: Spree.pathFor('solidus_paypal_braintree/transactions')
     },
-
+
```
tvdeyen: unnecessary whitespace sneaked into here :)

app/assets/javascripts/solidus_paypal_braintree/braintree_errors.js.erb
```diff
@@ -0,0 +1,17 @@
+BraintreeError = {
```
tvdeyen: Sorry, just saw that this is a `js.erb` file instead of an html partial. Although not very obviously, but `js.erb` files are very problematic, because changes to the translations won't trigger this file to be re-compiled on next asset compilation. Please make this is html.erb file that gets rendered inside the payment form.

Gamabunta94: Thank you for the feedback.
I changed this behavior and now it should work fine.

kennyadsl:
@tvdeyen can we merge this one? I think it's ok but I don't have permissions to merge here.

tvdeyen:
@kennyadsl merged

kennyadsl:
Thanks!

PR #136, opened 2017-11-21
Add a partial for the hash preference field

### LINKED ISSUE #132: Missing partial spree/admin/shared/preference_fields/_hash
I just installed the gem and follow the documentation to add a new payment method, but when I first save the method I get this error in the admin:
```
Missing partial spree/admin/shared/preference_fields/_hash
```
Yes, the partial is missing. I don't know how to solve this.

Latest gem and solidus v2.4.0

tvdeyen:
We store two settings inside a hash preference field. Solidus 2.4
is missing this partial. As will hopefully be fixed with Solidus 2.5
we load our own partial if we are using Solidus 2.4

Closes #132

### DIFF
--- a/lib/solidus_paypal_braintree.rb
+++ b/lib/solidus_paypal_braintree.rb
@@ -1,5 +1,4 @@
 require 'solidus_core'
-require 'solidus_support'
 require 'solidus_paypal_braintree/engine'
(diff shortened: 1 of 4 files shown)


tvdeyen:
@adammathys sorry, should have checked before bothering you. Should be good now.

PR #142, opened 2017-12-04
Recalculate payment total when order total decreases

dholdren:
- Apple Pay does not send street address until the last step. If street
  is a factor in the tax calculation you will have a discrepency after
  the payment is made. The payment amount wasn't decreased when
  the order total was lower


dholdren:
I added a spec

tvdeyen:
@dholdren could you please address the Rubocop offenses? Unfortunately it's part of our Travis builds and thats why its failing. 

NOTE TO SELF: Use HoundCI instead of Travis CI for Rubocop.

dholdren:
@tvdeyen sorry for the long delay, the rubocop offenses have been addressed

PR #143, opened 2017-12-05
Use remote transaction status for admin actions

tvdeyen:
Solidus usually uses the payment state for deciding if a certain admin action (capture, credit, void) is possible for a payment. Do to the async nature of the web this might get out of sync and leads to payments in a non-actionable state for admins. By asking the Braintree gateway for the current transaction status we know if a certain action is possible or not.

I also sneaked in a change that makes it possible for other payment types (besides credit cards) to access the payment method from the gateway. It doesn't make sense to not allow to read the data for other payment types (PayPal, ApplePay) which we support.

### DIFF
--- a/app/models/solidus_paypal_braintree/gateway.rb
+++ b/app/models/solidus_paypal_braintree/gateway.rb
@@ -25,6 +25,10 @@ module SolidusPaypalBraintree
       Braintree::Transaction::Status::Settled
     ].freeze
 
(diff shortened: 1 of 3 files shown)


tvdeyen:
On second thought this might not be a good idea, because the API could be down and this would cause time outs whenever someone enters the admin payments screen. In order to get the correct state if the state gets out of sync we should add a button to retrieve the latest state instead.

PR #144, opened 2017-12-06
Restart checkout when cart Paypal button is used

alepore:
The problem:
We have two different Paypal flows here: the express checkout from the cart (address comes from Paypal) and the normal Solidus checkout (address comes from Solidus).

If an user starts the express checkout, and then comes back to the cart (from the confirm page) and restarts it with a different address, we currently don't recalculate all the address-dependent order stuff, thus we can end up with wrong taxes and shipping methods.

This is a quick solution for that: the cart button sends a special param to the Solidus controller that restarts the checkout flow.
The normal checkout flow is unaltered.

Comments or better solutions are welcome :)


alepore:
@luukveenis @adammathys what do you think about this? :)

PR #164, opened 2018-04-20
Update PayPal configuration instructions

kennyadsl:
Add needed PayPal configuration to make this extension work into the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,6 +187,19 @@ A PayPal button can also be included on the cart view to enable express checkout
 render "spree/shared/paypal_cart_button"
 ```
 
+#### PayPal configuration
+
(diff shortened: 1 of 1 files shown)


kennyadsl:
Yes, it worked for me. Anyway I'm ok removing the validation since we've spent something like one day to understand that this option was causing the issue. 

I'd leave the README change anyway, as instructions about how to let PayPal return the phone back to Solidus.

If you do agree I can change the last sentence and rebase against master, let me know!

alepore:
@kennyadsl i tried again with a new sandbox and this seems to work, so yeah, i would merge this because it's important. A plain solidus install will not work without this setup.

kennyadsl:
@tvdeyen done!

alepore:
@kennyadsl rebase pls

kennyadsl:
Sad sad

PR #168, opened 2018-05-18
Express Checkout support

aldesantis:
Allows users to choose between Vault and Express Checkout via a gateway-level preference.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,8 @@ Payment methods can accept preferences either directly entered in admin, or from
         environment: Rails.env.production? ? 'production' : 'sandbox',
         merchant_id: ENV['BRAINTREE_MERCHANT_ID'],
         public_key: ENV['BRAINTREE_PUBLIC_KEY'],
-        private_key: ENV['BRAINTREE_PRIVATE_KEY']
+        private_key: ENV['BRAINTREE_PRIVATE_KEY'],
+        paypal_flow: 'vault', # 'checkout' is accepted too
       }
     )
   end
@@ -155,8 +156,10 @@ store's configuration.
 The checkout view
 [initializes the PayPal button](/lib/views/frontend/spree/checkout/payment/_paypal_braintree.html.erb)
 using the
-[vault flow](https://developers.braintreepayments.com/guides/paypal/overview/javascript/v3),
-which allows the source to be reused.
+[Vault flow](https://developers.braintreepayments.com/guides/paypal/overview/javascript/v3),
+which allows the source to be reused. If you want, you can use [Checkout with PayPal](https://developers.braintreepayments.com/guides/paypal/checkout-with-paypal/javascript/v3)
+instead, which doesn't allow you to reuse sources but allows your customers to pay with their PayPal
+balance (see setup instructions).
 
 If you are creating your own checkout view or would like to customize the
 [options that get passed to tokenize](https://braintree.github.io/braintree-web/3.6.3/PayPal.html#tokenize)
--- a/app/models/solidus_paypal_braintree/gateway.rb
+++ b/app/models/solidus_paypal_braintree/gateway.rb
@@ -264,6 +264,12 @@ module SolidusPaypalBraintree
 
     private
 
+    # Whether to store this payment method in the PayPal Vault. This only works when the checkout
+    # flow is "vault", so make sure to call +super+ if you override it.
+    def store_in_vault
+      preferred_paypal_flow == 'vault'
+    end
+
     def logger
       Braintree::Configuration.logger.clone.tap do |logger|
         logger.level = Rails.logger.level
@@ -295,7 +301,7 @@ module SolidusPaypalBraintree
       end
 
       params[:channel] = "Solidus"
-      params[:options] = { store_in_vault_on_success: (preferred_paypal_flow == 'vault') }
+      params[:options] = { store_in_vault_on_success: store_in_vault }
 
       if submit_for_settlement
         params[:options][:submit_for_settlement] = true
@@ -366,7 +372,7 @@ module SolidusPaypalBraintree
     def customer_profile_params(payment)
       params = {}
 
-      if preferred_paypal_flow == 'vault' && payment.source.try(:nonce)
+      if store_in_vault && payment.source.try(:nonce)
         params[:payment_method_nonce] = payment.source.nonce
       end
 
--- a/app/models/solidus_paypal_braintree/gateway.rb
+++ b/app/models/solidus_paypal_braintree/gateway.rb
@@ -38,6 +38,9 @@ module SolidusPaypalBraintree
     preference(:merchant_currency_map, :hash, default: {})
     preference(:paypal_payee_email_map, :hash, default: {})
 
+    # Which checkout flow to use (vault/checkout)
+    preference(:paypal_flow, :string, default: 'vault')
+
     def partial_name
       "paypal_braintree"
     end
@@ -292,7 +295,7 @@ module SolidusPaypalBraintree
       end
 
       params[:channel] = "Solidus"
-      params[:options] = { store_in_vault_on_success: true }
+      params[:options] = { store_in_vault_on_success: (preferred_paypal_flow == 'vault') }
 
       if submit_for_settlement
         params[:options][:submit_for_settlement] = true
@@ -363,7 +366,7 @@ module SolidusPaypalBraintree
     def customer_profile_params(payment)
       params = {}
 
-      if payment.source.try(:nonce)
+      if preferred_paypal_flow == 'vault' && payment.source.try(:nonce)
         params[:payment_method_nonce] = payment.source.nonce
       end
 
--- a/app/views/spree/shared/_paypal_checkout_button.html.erb
+++ b/app/views/spree/shared/_paypal_checkout_button.html.erb
@@ -15,7 +15,9 @@
   }
 
   var paypalOptions = {
-    flow: 'vault',
+    flow: '<%= SolidusPaypalBraintree::Gateway.first.preferred_paypal_flow %>',
+    amount: '<%= current_order.total %>',
+    currency: '<%= current_order.currency %>',
     enableShippingAddress: true,
     shippingAddressOverride: address,
     shippingAddressEditable: false,


README.md
```diff
@@ -46,7 +46,8 @@ Payment methods can accept preferences either directly entered in admin, or from
         environment: Rails.env.production? ? 'production' : 'sandbox',
         merchant_id: ENV['BRAINTREE_MERCHANT_ID'],
         public_key: ENV['BRAINTREE_PUBLIC_KEY'],
-        private_key: ENV['BRAINTREE_PRIVATE_KEY']
+        private_key: ENV['BRAINTREE_PRIVATE_KEY'],
+        paypal_flow: 'vault', # 'checkout' is accepted too
```
kennyadsl: isn't `checkout_with_paypal` in the comment?

aldesantis: @kennyadsl updated.

app/assets/javascripts/spree/frontend/paypal_button.js
```diff
@@ -19,7 +19,7 @@ $(document).ready(function() {
       }).
       load(function() {
         var paypalOptions = {
-          flow: 'vault',
+          flow: window.payPalFlow,
```
tvdeyen: This button is rendered in the cart. Are we sure that the global variable is set?

lib/views/frontend/spree/checkout/payment/_paypal_braintree.html.erb
```diff
@@ -2,6 +2,10 @@
 <% id = payment_method.id %>
 
 <% content_for :head do %>
+  <script>
+    window.payPalFlow = '<%= SolidusPaypalBraintree::Gateway.first.preferred_paypal_flow %>';
```
tvdeyen: Are we sure that this global variable is available if we render the button in the cart?

app/assets/javascripts/spree/frontend/paypal_button.js
aldesantis: @tvdeyen I agree that this is a bit brittle. Do you have any suggestions?

tvdeyen: Not without refactoring how we setup the paypal button in the cart. Ideally we would do it like we do it in checkout payment. Loading the script and setting in up in a view partial.

aldesantis: @tvdeyen so something like this, right?

```erb
<%= render 'paypal_button' %>
```

And `_paypal_button.html.erb` would be just:

```erb
<script>
$(document).ready(function() {
  if (document.getElementById("empty-cart")) {
    $.when(
      $.getScript("https://js.braintreegateway.com/web/3.22.1/js/client.min.js"),
      $.getScript("https://js.braintreegateway.com/web/3.22.1/js/paypal.min.js"),
      $.getScript("https://js.braintreegateway.com/web/3.22.1/js/data-collector.min.js")
    ).done(function() {
      $('<script/>').attr({
        'data-merchant' : "braintree",
        'data-id' : "paypal-button",
        'data-button' : "checkout",
        'data-color' : "blue",
        'data-size' : "medium",
        'data-shape' : "pill",
        'data-button_type' : "button",
        'data-button_disabled' : "true"
      }).
      load(function() {
        var paypalOptions = {
          flow: '<%= SolidusPaypalBraintree::Gateway.first.preferred_paypal_flow %>',
          enableShippingAddress: true
        }
        var button = new SolidusPaypalBraintree.createPaypalButton(document.querySelector("#paypal-button"), paypalOptions);
        return button.initialize();
      }).
      insertAfter("#content").
      attr('src', 'https://www.paypalobjects.com/api/button.js?')
    });
  }
});
</script>
```
I can go ahead and implement it, but I'm afraid this might be a breaking change? I guess we _could_ keep the original asset for backwards compatibility and implement the partial for when you want to tweak the flow, but I'm not sure this is a good idea. What do you think?

tvdeyen: Thanks, yes. Also taken care of in #170

tvdeyen:
I would love to merge #170 before merging this.

aldesantis:
@tvdeyen makes sense. Just ping me once #170 is merged and I'll be happy to update this one. :)

alepore:
🚨 @aldesantis #170 was merged.
I moved the paypal cart button on a partial (which is mostly an example of how to use it). 
that's a breaking change, yes, but the old one was not so good, i would just write in on the changelog.

alepore:
can someone restart the failing spec?

tvdeyen:
@alepore restarted. Fingers crossed

aldesantis:
Hey @tvdeyen any ideas as to why the tests might be failing?

alepore:
@aldesantis please rebase on master!

aldesantis:
@alepore done! :)

alepore:
aaand random failures are back (but seems less now) 
@tvdeyen please restart https://travis-ci.org/solidusio/solidus_paypal_braintree/jobs/403187947 🙏

aldesantis:
@alepore this looks good now!

app/models/solidus_paypal_braintree/gateway.rb
```diff
@@ -292,7 +295,7 @@ def transaction_options(source, options, submit_for_settlement = false)
       end
 
       params[:channel] = "Solidus"
-      params[:options] = { store_in_vault_on_success: true }
+      params[:options] = { store_in_vault_on_success: (preferred_paypal_flow == 'vault') }
```
tvdeyen: If we already make this change (what I think is great), we may want to extract that into a method, so stores don’t have to override this whole method if they don’t want to store the payment source, regardless of the pypal flow.

aldesantis: @tvdeyen should we make this a preference on the gateway perhaps?

tvdeyen: The problem is that non devs are able to set preferences in the admin and store-in-vault only works if the flow is also vault.

tvdeyen: And ideally the customer should be able to choose. Neither the dev nor the admin, the user needs to be able to decide if she wants to store a token thats able to withdraw money without further verification.

aldesantis: @tvdeyen agreed, it shouldn't be a preference. If it's configurable by the devs though should it be a method or some other kind of internal configuration option? For instance:
```ruby
SolidusPaypalBraintree.store_sources = true
```
I think this is better than the method, because theoretically methods are not a "clean" extension point - they have to be overridden with `prepend` or `class_eval`.

Re: letting users choose if they want to store the token, I think that's a good additional feature for later. How would you go about it? Some column on the order/payment model?

aldesantis: @tvdeyen bump? :)

tvdeyen: > I think this is better than the method, because theoretically methods are not a "clean" extension point - they have to be overridden with `prepend` or `class_eval`.

@aldesantis a method is the only way I can imagine this will work with core Solidus right know as we need to know if the user wants to store a certain payment source.

tvdeyen: > Re: letting users choose if they want to store the token, I think that's a good additional feature for later. How would you go about it? Some column on the order/payment model?

@aldesantis yes a boolean on the payment source class toggled by a checkbox in the checkout view

aldesantis:
@tvdeyen added a `store_in_vault` method. Let me know if this is what you were looking for.

app/models/solidus_paypal_braintree/gateway.rb
```diff
@@ -292,7 +301,9 @@ def transaction_options(source, options, submit_for_settlement = false)
       end
 
       params[:channel] = "Solidus"
-      params[:options] = { store_in_vault_on_success: true }
+      params[:options] = {
+        store_in_vault_on_success: (store_in_vault && preferred_paypal_flow == 'vault')
```
tvdeyen: Could we move this `preferred_paypal_flow == 'vault' ` into the `store_in_vault` method? I think it makes more sense in there.

app/models/solidus_paypal_braintree/gateway.rb
```diff
@@ -363,7 +374,7 @@ def paypal_payee_email_for(source, options)
     def customer_profile_params(payment)
       params = {}
 
-      if payment.source.try(:nonce)
+      if preferred_paypal_flow == 'vault' && payment.source.try(:nonce)
```
tvdeyen: Could we also use the `store_in_vault` method here?

app/models/solidus_paypal_braintree/gateway.rb
aldesantis: @tvdeyen the problem is that `store_in_vault_on_success: true` will fail if the flow is not `vault`, so I wanted to avoid potential issues with that.

app/models/solidus_paypal_braintree/gateway.rb
aldesantis: @tvdeyen I think this is needed even when you don't store the payment source.

aldesantis: @tvdeyen sorry, I misread your initial comment. Will correct. :)

app/views/spree/shared/_paypal_checkout_button.html.erb
```diff
@@ -15,7 +15,9 @@
   }
 
   var paypalOptions = {
-    flow: 'vault',
+    flow: '<%= SolidusPaypalBraintree::Gateway.first.preferred_paypal_flow %>',
```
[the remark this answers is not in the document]
aldesantis: @tvdeyen no, because you might use the Vault flow without storing the payment method in the vault.

app/models/solidus_paypal_braintree/gateway.rb
aldesantis: @tvdeyen fixed.

aldesantis:
@tvdeyen should be ready for review again.

app/views/spree/shared/_paypal_checkout_button.html.erb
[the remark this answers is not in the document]
tvdeyen: The whole purpose of the `vault` flow is to store payment methods for later re-use, no? Why should one use the `vault` flow if the payment method must not be stored in it? Do you have an example of that use case?

app/models/solidus_paypal_braintree/gateway.rb
```diff
@@ -292,7 +301,9 @@ def transaction_options(source, options, submit_for_settlement = false)
       end
 
       params[:channel] = "Solidus"
-      params[:options] = { store_in_vault_on_success: true }
+      params[:options] = {
+        store_in_vault_on_success: store_in_vault
+      }
```
tvdeyen: I think this can now be one line again

app/views/spree/shared/_paypal_checkout_button.html.erb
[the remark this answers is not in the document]
aldesantis: @tvdeyen I'm not sure about the use case, but there are two points in favor of keeping these separate in my opinion:

- We can't know the return value of `store_in_vault` _before_ we start the PayPal flow.
- PayPal supports using the Vault flow without storing the payment source, so I don't think we should prevent users from doing this, especially since it comes at no additional cost for us.

aldesantis:
@tvdeyen I see you mentioned one last question blocking this but can't find it, could you repeat it here please?

PR #170, opened 2018-05-30
Update Braintree and PayPal libs

alepore:
- Update Braintree libs to v3.34.0
- Update the PayPal integration to use the new and supported version (checkout.js)
- Move PayPal and Apple Pay buttons into partials for easier use

Other things included:

- Feature specs runs on headless chrome instead of poltergeist

Breaking changes:

- The cart button is not automatically shown on the frontend anymore.
I think using using a partial is a better idea here. The current JS only works with the default solidus_frontend anyway and seems pretty hacky. (We could keep the old default behaviour using `deface`)


alepore:
Please restart the failed builds, i don't have the power :)

PR #173, opened 2018-07-05
Fix Gateway#try_void expecting a respose code instead of a payment

aldesantis:
Solidus passes a Spree::Payment object to Gateway#try_void, but we are expecting a response code right now.


alepore:
@tvdeyen pls restart https://travis-ci.org/solidusio/solidus_paypal_braintree/jobs/403187114 🙏

PR #174, opened 2018-07-06
Fix error display

alepore:
We currently don't display transaction errors on the checkout. This behaviour is very confusing for the user.
Even if the user may not be able to fix the error by itself, we should make clear that something bad happened.

This is the result: 
screen shot 2018-07-06 at 16 06 55

### DIFF
--- a/app/assets/javascripts/solidus_paypal_braintree/paypal_button.js
+++ b/app/assets/javascripts/solidus_paypal_braintree/paypal_button.js
@@ -70,7 +70,22 @@ SolidusPaypalBraintree.PaypalButton.prototype._tokenizeCallback = function(token
       window.location.href = response.redirectUrl;
     },
     error: function(xhr) {
-      console.error("Error submitting transaction");
+      var errorText = BraintreeError.DEFAULT;
+
+      if (xhr.responseJSON && xhr.responseJSON.errors) {
+        var errors = [];
+        $.each(xhr.responseJSON.errors, function(key, values) {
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/solidus_paypal_braintree/paypal_button.js
```diff
@@ -70,7 +70,16 @@ SolidusPaypalBraintree.PaypalButton.prototype._tokenizeCallback = function(token
...
+      var errors = []
+      $.each(xhr.responseJSON.errors, function(key, values) {
+        $.each(values, function(index, value) {
+          errors.push(key + " " + value)
+        });
+      });
+      var errorText = errors.join(", ");
+
+      console.error("Error submitting transaction: " + errorText);
+      SolidusPaypalBraintree.showError(BraintreeError.DEFAULT + " " + errorText);
```
tvdeyen: A validation error is not "Bad", so we should not prepend all errors with the default message.

alepore:
@tvdeyen 👍 i agree.
I removed the prefix, and made the default message the fallback.

tvdeyen:
@alepore thanks. Build is failing because of rubocop. 😢 

We really should not run rubocop on Travis CI, but use Hound for this.

tvdeyen:
@alepore https://github.com/solidusio/solidus_paypal_braintree/pull/176

tvdeyen:
@alepore one final rebase we are good to go. Thanks

PR #176, opened 2018-07-10
Do not run Rubocop on each spec run

tvdeyen:
All our travis ci builds fail because rubocop is unhappy with something. The build should not fail because of syntax issues. And we should use HoundCI for this instead.


alepore:
Is this suite cursed or something?

tvdeyen:
> Is this suite cursed or something?

Yes

alepore:
This is already on master and can be closed

PR #177, opened 2018-07-10
Update CI Ruby and specs config

alepore:
Any reason why we use the oldest supported Ruby version?


alepore:
@tvdeyen 🍏 💚 📗

tvdeyen:
Thanks

PR #190, opened 2018-10-16
Fix test gems versions

tvdeyen:
Removes gems we do not need anymore as we do not test Solidus < 2.2 any more and use `factory_bot` versions that work with Solidus branches.


kennyadsl:
Closed with #193

PR #201, opened 2019-01-03
Check for rails-ujs object when re-enabling button

masonjeffreys:
When using this gem in a project with Rails 5.2.0, I noticed that the "Save and Continue" button to submit credit card information using the Braintree hosted fields was disabled.

I tracked the issue to a call to $.rails.enableFormElement. This method likely exists for people that are using older versions of Rails and jquery-ujs. However, at Rails 5.1.0, jquery-ujs was "adapted" into the rails core as rails-ujs. rails-ujs has a slightly different way of enabling elements.

I added a check to make sure that if $.rails exists, it has the appropriate method. If not, a further check is performed to determine if Rails (the rails-ujs object) exists. If so, it uses a different way to enable the submit button.

Let me know if you see anything I can improve.
Thanks,
Jeff

### DIFF
--- a/app/assets/javascripts/solidus_paypal_braintree/checkout.js
+++ b/app/assets/javascripts/solidus_paypal_braintree/checkout.js
@@ -15,9 +15,16 @@ $(function() {
      * https://github.com/rails/jquery-rails/blob/master/vendor/assets/javascripts/jquery_ujs.js#L517
      * The only way we can re-enable it is by delaying longer than that timeout
      * or stopping propagation so their submit handler doesn't run. */
-    if ($.rails) {
+    if ($.rails && typeof $.rails.enableFormElement !== 'undefined') {
       setTimeout(function () {
         $.rails.enableFormElement($submitButton);
+        $submitButton.attr("disabled", false).removeClass("disabled").addClass("primary");
+      }, 100);
+    } else if (typeof Rails !== 'undefined' && typeof Rails.enableElement !== 'undefined') {
+      /* Indicates that we have rails-ujs instead of jquery-ujs. Rails-ujs was added to rails
+       * core in Rails 5.1.0 */
+      setTimeout(function () {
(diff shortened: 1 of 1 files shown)


masonjeffreys:
#202

masonjeffreys:
Hmm...I don't know actually. That's interesting. Perhaps the problem is a bit more nuanced than I suspected. I see the error when I'm using a custom front end for solidus. So maybe there is something in the custom frontend that is bringing this issue to light?

I'll try to do a fresh solidus build and see if I can recreate the issue.

masonjeffreys:
Okay I have an idea here. Looks like the dummy app is including jquery_ujs: solidus_paypal_braintree/spec/dummy/vendor/assets/javascripts/spree/frontend/all.js
`//`
`//= require jquery`
`//= require jquery_ujs`

If I change _require jquery_ujs_ to _require rails-ujs_, then we get two errors in the spec:
`rspec ./spec/features/frontend/braintree_credit_card_checkout_spec.rb:41 # entering credit card details with valid credit card data checks out successfully`
`rspec ./spec/features/frontend/braintree_credit_card_checkout_spec.rb:84 # entering credit card details with invalid credit card data user enters valid data allows them to resubmit and complete the purchase`

`Failure/Error: click_button("Save and Continue")`
     
     'Capybara::ElementNotFound:'
       'Unable to find visible button "Save and Continue"'

After I implement the fix in this PR, all tests pass.

**So I believe this error will show up if you are using Rails 5.1.0 or greater and do not include the spree/frontend/all.js file which will include the jquery_ujs file.**

masonjeffreys:
Awesome. Thanks @kennyadsl !

spaghetticode:
@masonjeffreys thanks for the fix 👏 

I think this is the same problem we're having now in master (I'm seeing `$.rails.enableFormElement is not a function` in a JS console of a failing spec), as now @kennyadsl PR on Solidus is merged.

Issue #208, opened 2019-01-29
Ambiguous behavior for generate token method


skukx:
The `SolidusPaypalBraintree::Gateway#generate_token` method has ambiguous behavior. If token generation is disabled it returns a string acting as a "[error message](https://github.com/solidusio/solidus_paypal_braintree/blob/master/app/models/solidus_paypal_braintree/gateway.rb#L219)" as described in the documentation. If the generation is enabled it also returns a string. 

https://github.com/solidusio/solidus_paypal_braintree/blob/master/app/models/solidus_paypal_braintree/gateway.rb#L236

Rather than returning a string which is also expected in on success, possibly raise an error with that string as a message. Then the app can handle that error in a way it sees fit. Otherwise it is not clear whether gernation is enabled or disabled unless using string comparison at every check

PR #209, opened 2019-01-29
Fix ambiguous behavior in generate_token method

skukx:
The `SolidusPaypalBraintree::Gateway#generate_token` method is ambiguous
on success and failure. Both results will return a string. To be more
explicit, this method should raise an error which can be handled by the
app.

Closes https://github.com/solidusio/solidus_paypal_braintree/issues/208

### DIFF
--- a/app/controllers/solidus_paypal_braintree/client_tokens_controller.rb
+++ b/app/controllers/solidus_paypal_braintree/client_tokens_controller.rb
@@ -7,7 +7,7 @@ module SolidusPaypalBraintree
     before_action :load_gateway
 
     def create
-      render json: { client_token: @gateway.generate_token, payment_method_id: @gateway.id }
+      render json: { client_token: generate_token, payment_method_id: @gateway.id }
     end
 
     private
@@ -24,6 +24,13 @@ module SolidusPaypalBraintree
           end
         @gateway = ::SolidusPaypalBraintree::Gateway.where(active: true).merge(store_payment_methods_scope).first!
       end
+    end
+
+    def generate_token
+      @gateway.generate_token
+    rescue ::SolidusPaypalBraintree::Gateway::TokenGenerationDisabledError => error
+      Rails.logger.error error
+      nil
     end
   end
 end
--- a/app/controllers/solidus_paypal_braintree/client_tokens_controller.rb
+++ b/app/controllers/solidus_paypal_braintree/client_tokens_controller.rb
@@ -7,7 +7,12 @@ module SolidusPaypalBraintree
     before_action :load_gateway
 
     def create
-      render json: { client_token: generate_token, payment_method_id: @gateway.id }
+      token = @gateway.generate_token
+      if token
+        render json: { client_token: token, payment_method_id: @gateway.id }
+      else
+        render json: { error: Gateway::TOKEN_GENERATION_DISABLED_MESSAGE }, status: :unprocessable_entity
+      end
     end
 
     private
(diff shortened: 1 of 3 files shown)


app/models/solidus_paypal_braintree/gateway.rb
```diff
@@ -4,6 +4,9 @@ module SolidusPaypalBraintree
   class Gateway < ::Spree::PaymentMethod
     include RequestProtection
 
+    class Error < StandardError; end
```
ericsaupe: Is the `Error` class needed? `TokenGenerationDisabledError` could inherit from `StandardError`

skukx: I'm not sure how many errors are planned or could be used. We could but sometimes its nice to just rescue `Gateway::Error` which catches all errors within the class

kennyadsl: Are there any cons in adding this `Error` class later, only if needed?

app/controllers/solidus_paypal_braintree/client_tokens_controller.rb
```diff
@@ -18,5 +18,12 @@ def load_gateway
         @gateway = ::SolidusPaypalBraintree::Gateway.where(active: true).merge(store_payment_methods_scope).first!
       end
     end
+
+    def generate_token
+      @gateway.generate_token
+    rescue ::SolidusPaypalBraintree::Gateway::TokenGenerationDisabledError => error
+      Rails.logger.warn error
```
kennyadsl: Shouldn't be `Rails.logger.error`?

app/models/solidus_paypal_braintree/gateway.rb
skukx: Nope, I can do that

app/controllers/solidus_paypal_braintree/client_tokens_controller.rb
skukx: I wasn't sure what level it should be. I'm ok changing it to `Rails.logger.error`

kennyadsl:
@skukx thanks! Last thing: I'd also squash everything in a single commit, what do you think?

skukx:
@ericsaupe I'm curious if `force pushing` affects travis CI. Getting those random failures again

app/controllers/solidus_paypal_braintree/client_tokens_controller.rb
```diff
@@ -5,7 +5,7 @@ class ClientTokensController < Spree::Api::BaseController
     before_action :load_gateway
 
     def create
-      render json: { client_token: @gateway.generate_token, payment_method_id: @gateway.id }
+      render json: { client_token: generate_token, payment_method_id: @gateway.id }
```
tvdeyen: This could be rewritten as

```suggestion
      token = @gateway.generate_token
      if token
        render json: { client_token: generate_token, payment_method_id: @gateway.id }
      else
        render json: { error: Gateway::TOKEN_GENERATION_DISABLED_MESSAGE }, status: 422
      end
```

tvdeyen:
> @ericsaupe I'm curious if `force pushing` affects travis CI. Getting those random failures again

Yes, force pushes trigger a new build on Travis. The random failures are because of the nature of the "not so ideal written" integrations tests.

seand7565:
@skukx Hey there, from the future! I would like to merge this in soon. Do you have a moment to give this a rebase? Thanks!

skukx:
> @skukx Hey there, from the future! I would like to merge this in soon. Do you have a moment to give this a rebase? Thanks!

@seand7565 Yes, I'll do that asap

seand7565:
> > @skukx Hey there, from the future! I would like to merge this in soon. Do you have a moment to give this a rebase? Thanks!
> 
> @seand7565 Yes, I'll do that asap

Fantastic, thanks! Looks like there's just two small linting issues and this should be good to go.

kennyadsl:
This is another breaking change, people relying on that specific response even for failing cases. We will definitely need to release a new version with a clear changelog for this issue.

tvdeyen (approved):
Code wise 👌 

I agree with Alberto that this is a breaking change, but in a good sense. The former behavior was just wrong.

kennyadsl (approved):

PR #211, opened 2019-01-29
Add association from user to customer

### LINKED ISSUE #210: Add Association for spree user to customer
An association exists from `SolidusPaypalBraintree::Customer` to `Spree.user_class` however there is no  association on the user to the braintree customer.

```ruby
Spree.user_class.class_eval do
  has_one :braintree_customer, class_name: 'SolidusPaypalBraintree::Customer'
end
```

skukx:
Closes #210

### DIFF
--- a/app/models/spree/user_decorator.rb
+++ b/app/models/spree/user_decorator.rb
@@ -1,0 +1,4 @@
+Spree.user_class.class_eval do
+  has_one :braintree_customer, class_name: 'SolidusPaypalBraintree::Customer',
+                               inverse_of: :user
+end


app/models/spree/user_decorator.rb
```diff
@@ -0,0 +1,4 @@
+Spree.user_class.class_eval do
+  has_one :braintree_customer, class_name: 'SolidusPaypalBraintree::Customer',
```
ericsaupe: This might benefit from having an `index` as well. Are we ever finding a user by this field?

skukx: I think there may be an index already on the `user_id` column of the `SolidusPaypalBraintree::Customer`

PR #212, opened 2019-01-30
one to one relationship for gateway customer

### LINKED ISSUE #206: New Braintree Customer created for every payment method created
Similar to this issue here: https://github.com/solidusio/solidus_braintree/issues/56

When creating a new payment method, `@gateway.generate_token` will create a nonce for a brand new customer. We should try to keep a one to one relationship between Braintree customers and Spree Users.

This may require another field or table to keep track of the gateway customer id. Then when requesting a token from braintree something like the following should be called if a user is logged in.
```ruby
@gateway.generate_token customer_id: user.gateway_customer_id
```

This ensures that the new payment method is added to the same braintree customer as all others in the past.

skukx:
Ensure that when creating braintree customers we create one per spree
user. To do this, first generate a token with the given customer id if
available. When a braintree customer already exists, rather than create
a new braintree customer, add the payment method to the existing
braintree customer.

Closes #206

### DIFF
--- a/app/controllers/solidus_paypal_braintree/client_tokens_controller.rb
+++ b/app/controllers/solidus_paypal_braintree/client_tokens_controller.rb
@@ -1,11 +1,9 @@
 module SolidusPaypalBraintree
-  class ClientTokensController < Spree::Api::BaseController
-    skip_before_action :authenticate_user
-
+  class ClientTokensController < Spree::BaseController
     before_action :load_gateway
 
     def create
-      render json: { client_token: @gateway.generate_token, payment_method_id: @gateway.id }
+      render json: { client_token: generate_token, payment_method_id: @gateway.id }
     end
 
     private
@@ -17,6 +15,21 @@ module SolidusPaypalBraintree
         store_payment_methods_scope = current_store.payment_methods.empty? ? ::SolidusPaypalBraintree::Gateway.all : ::SolidusPaypalBraintree::Gateway.where(id: current_store.payment_method_ids)
         @gateway = ::SolidusPaypalBraintree::Gateway.where(active: true).merge(store_payment_methods_scope).first!
       end
+    end
+
+    def generate_token
+      options = {}
+      options[:customer_id] = customer_id if customer_id.present?
+
+      @gateway.generate_token(options)
+    rescue ::SolidusPaypalBraintree::Gateway::TokenGenerationDisabledError => error
+      Rails.logger.error error
+      nil
+    end
+
+    def customer_id
+      return unless try_spree_current_user&.braintree_customer
+      try_spree_current_user.braintree_customer.braintree_customer_id
     end
   end
 end
--- a/app/models/solidus_paypal_braintree/gateway.rb
+++ b/app/models/solidus_paypal_braintree/gateway.rb
@@ -4,6 +4,8 @@ module SolidusPaypalBraintree
   class Gateway < ::Spree::PaymentMethod
     include RequestProtection
 
+    class TokenGenerationDisabledError < StandardError; end
+
     # Error message from Braintree that gets returned by a non voidable transaction
     NON_VOIDABLE_STATUS_ERROR_REGEXP = /can only be voided if status is authorized/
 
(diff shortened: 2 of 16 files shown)


skukx:
@ericsaupe  https://github.com/solidusio/solidus_paypal_braintree/pull/212/files#diff-285f84b37d3de3e6ec6c226e28dc8a6aR222

app/models/solidus_paypal_braintree/customer.rb
```diff
@@ -1,4 +1,7 @@
 class SolidusPaypalBraintree::Customer < ApplicationRecord
   belongs_to :user, class_name: Spree::UserClassHandle.new
   has_many :sources, class_name: "SolidusPaypalBraintree::Source", inverse_of: :customer
+
+  validates :user_id, presence: true
+  validates :braintree_customer_id, presence: true
```
kennyadsl: What about existing records? Could we end up saving records previously created that now are no more valid and can't be updated anymore?

solidus_paypal_braintree.gemspec
```diff
@@ -22,6 +22,7 @@ Gem::Specification.new do |s|
   s.add_dependency "braintree", '~> 2.65'
   s.add_dependency 'activemerchant', '~> 1.48'
 
+  s.add_development_dependency 'byebug'
```
kennyadsl: I'd move this byebug change into another PR, or at least a separate commit

app/models/solidus_paypal_braintree/gateway.rb
```diff
@@ -375,5 +391,16 @@ def customer_profile_params(payment)
 
       params
     end
+
+    def payment_method_params(payment)
```
kennyadsl: This method smells like it shouldn't be here, what about a separate class? Anyway I'd call this method `payment_method_attributes_from(payment)` at least or `payment_to_braintree_payment_method_attributes(payment)`, not sure what's best, or worst? 🙂

app/models/solidus_paypal_braintree/customer.rb
skukx: Yeah, this has been removed

spec/spec_helper.rb
```diff
@@ -56,6 +57,7 @@
   config.use_transactional_fixtures = false
   config.example_status_persistence_file_path = "tmp/failed_examples.txt"
 
+  config.include Devise::Test::ControllerHelpers, type: :controller
```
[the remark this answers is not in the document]
skukx: The controller inheritance has changed from `Spree::Api::BaseController` to `Spree::BaseController`. This was done because we need to generate the client token for braintree with the customer id. To get that we need to know the user logged in.

This change resulted in spec failures:
```
Devise::MissingWarden:
       Devise could not find the `Warden::Proxy` instance on your request environment.
```

app/models/solidus_paypal_braintree/gateway.rb
skukx: I agree with this, I followed suit with the `customer_profile_params(payment)` method for consistency.

skukx:
@kennyadsl The reason all the VCR cassettes changed is because there is now a different api call that is made.  When there is a braintree customer attached to a spree user then we `Add a payment method`. Otherwise we are `Creating a new customer` with the new payment method.

This is one of the reasons I'm not a fan of VCR (there are many). VCR is very quick and simple but can create issues over time. I would love to open a PR to remove VCR altogether and stub the service with sinatra.

As far as commits go, I'm probably going to squash all of this into 1 commit

skukx:
There is a bug with the `ClientTokensController`. The controller generates the token for the current user. However if that endpoint is called from admin then it will attach the payment source to the administrating user rather than the user on the order

app/controllers/solidus_paypal_braintree/client_tokens_controller.rb
```diff
@@ -18,5 +16,20 @@ def load_gateway
...
+      options[:customer_id] = customer_id if customer_id.present?
+
+      @gateway.generate_token(options)
+    rescue ::SolidusPaypalBraintree::Gateway::TokenGenerationDisabledError => error
+      Rails.logger.error error
+      nil
+    end
+
+    def customer_id
+      return unless try_spree_current_user&.braintree_customer
```
skukx: There is a bug with the `ClientTokensController`. The controller generates the token for the current user. However if that endpoint is called from admin then it will attach the payment source to the administrating user rather than the user on the order

kennyadsl: good catch!

skukx: Debating if the token should be generated when the view is rendered. Maybe as a hidden input

kennyadsl: Not sure, what's the difference?

skukx: See implementation 2 here: https://github.com/solidusio/solidus_braintree/pull/79

The main difference is that the view no longer makes an api request to obtain the client token but rather just generates it when the view renders.  The api request was only made if the new braintree payment method was selected (javascript events).  This change generates a token even if the user may not be adding a new payment method. Therefore it could mean a slight increase in api requests sent to braintree. I think this should be negligible though

kennyadsl: What about who is using this gem without a Rails frontend, eg on React or Vue.js? I think we still need this controller to work for them, right?

seand7565:
@skukx Hello again! Now that the other PR has been merged, would you mind rebasing this one as well? Thanks!

PR #214, opened 2019-02-04
Add byebug dependency

skukx:
Add byebug as a development dependency for debugging during development
and testing.


jacobherrington:
@skukx would you mind rebasing, just to get green specs? Obviously, this will pass just fine.

PR #216, opened 2019-02-07
Extension maintenance

aitbw:
This PR will introduce the following changes to get a green Travis build once again:

* Update Travis build matrix (see commit's description for details)
* Use dynamic attributes on factories
* Set `FactoryBot.use_parent_strategy` to `false` (see commit's description for details)
* Lock SQLite3 to version 1.3 (see commit's description for details)
* Fix deprecation warnings

PR #217, opened 2019-02-07
moved the paypal_braintree partial on the correct path

zener79:


kennyadsl:
@zener79 can you please explain why this is needed?

zener79:
@kennyadsl This is needed to avoid the error 
```
ActionView::Template::Error (Missing partial spree/api/payments/source_views/_paypal_braintree
```
fired on `solidus_api (2.8.0) app/views/spree/api/orders/_big.json.jbuilder:37`
```
    34:     payment_source = payment.source.is_a?(Spree::Payment) ? payment.source.source : payment.source
    35:                                                                               
    36:     if payment_source                                                            
    37:       json.partial!(                                                 
    38:         "spree/api/payments/source_views/#{payment.payment_method.partial_name}",
    39:         payment_source: payment_source                                         
    40:       )   
```

This error also prevents line_items from being shown in the order page on the backend as the ajax GET  request to `/api/orders/XXXXXXXX` returns a 500 status-code

kennyadsl:
Oh you are right, that file wasn't in the right position. I'd love to see a regression test for this. Maybe something very similar to what has been done in `solidus_affirm`: 

https://github.com/solidusio-contrib/solidus_affirm/blob/master/spec/requests/spree/api/orders_controller_spec.rb#L3-L29

Also, can you please against master now that specs have been fixed with #216?

PR #219, opened 2019-04-19
Fix translation key

jtapia:
Avoids scenarios where the translation is built like `spree.admin.tab.Braintree`, with the label this doesn't affect it

### DIFF
--- a/lib/solidus_paypal_braintree/engine.rb
+++ b/lib/solidus_paypal_braintree/engine.rb
@@ -54,6 +54,7 @@ module SolidusPaypalBraintree
           config.menu_items << config.class::MenuItem.new(
             [:braintree],
             'cc-paypal',
+            label: 'braintree',
             url: '/solidus_paypal_braintree/configurations/list',
             condition: -> { can?(:list, SolidusPaypalBraintree::Configuration) }
           )


kennyadsl:
@jtapia Agree with Thomas, thanks for the contribution, I'm sure it's something worth to be merged but can you please explain better why this is needed?

lib/solidus_paypal_braintree/engine.rb
```diff
@@ -54,6 +54,7 @@ def self.activate
           config.menu_items << config.class::MenuItem.new(
             [:braintree],
             'cc-paypal',
+            label: 'braintree',
```
kennyadsl: If a label is not specified, it will use the [first section](https://github.com/solidusio/solidus/blob/9ae2ed16bad7e29ea422fab1403118a3a0e66266/backend/app/models/spree/backend_configuration.rb#L58), which in this case is `:braintree`, that label [will be passed into the `tab` method](https://github.com/solidusio/solidus/blob/master/backend/app/views/spree/admin/shared/_tabs.html.erb#L7) which [just translate it](https://github.com/solidusio/solidus/blob/9ae2ed16bad7e29ea422fab1403118a3a0e66266/backend/app/helpers/spree/admin/navigation_helper.rb#L56). Not sure why this change is needed honestly.

PR #227, opened 2019-06-10
Stop importing address from PayPal

### LINKED ISSUE #226: On PayPal payment user shipping address gets set with uncorrect firstname and lastname values if they contain a space
When a user pays with Braintree PayPal (vault flow), the authorization callback gets parsed by the JavaScript code, and then submitted to server. The PayPal authorization callback provides the shipping address with a single field for the user lastname and firstname, `recipientName`. The client side code tries to infer the firstname and the lastname splitting this `recipientName` by space:

https://github.com/solidusio/solidus_paypal_braintree/blob/73e9a692148126745272f902a9ffece191b66225/app/assets/javascripts/solidus_paypal_braintree/paypal_button.js#L125-L126

The problem is that, if your name is for example `Maurizio De Santis`, the lastname will be set to `Maurizio` and the lastname to `De`, discarding the rest.

What's really wrong to me is that after that this shipping address value is submitted to the server, which updates the shipping address thus setting wrong values. But with the Braintree PayPal vault flow you can't change the shipping address via PayPal, so it's redundant to update it after the authorization call with the values that PayPal provides: the shipping address is the one that is already present on the server.

mdesantis:
Closes #226. When PayPal sends the authorization response to the server, it format the address with a single field for the first name and the last name, `recipientName`. We can’t safely infer the first name and the last name from that field (see #226), so we have to skip the user address updating. The good news is that the user address updating in this case is redundant, since Solidus always requires the address to be set before the payment happens, and such address is sent to PayPal with the `shippingAddressEditable` configuration set to `false`, so the address is necessarily the same as the one stored anyway.

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,18 +190,13 @@ A PayPal button can also be included on the cart view to enable express checkout
 render "spree/shared/paypal_cart_button"
 ```
 
-#### PayPal configuration
+#### PayPal
 
-If your store requires the [phone number into user addresses](https://github.com/solidusio/solidus/blob/859143f3f061de79cc1b385234599422b8ae8e21/core/app/models/spree/address.rb#L151-L153)
-you'll need to configure PayPal to return the phone back when it returns the
-address used by the user:
+##### Known limitations
 
-1. Log into your PayPal account
-2. Go to Profile -> My Selling Tools -> Website preferences
-3. Set Contact Telephone to `On (Required Field)` or `On (Optional Field)`
-
-Using the option `Off` will not make the address valid and will raise a
-validation error.
+It's not possible to let the user update the address via PayPal, because of
+storing formats incompatibilities between PayPal and Solidus; thus be sure that
+the PayPal client configuration has `shippingAddressEditable` set to `false`.
 
 ## Optional configuration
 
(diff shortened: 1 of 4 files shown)


tvdeyen:
I am not a fan of this. We should store the full name as `first_name` instead as Solidus has soft deprecated `last_name` a long time ago. 

The purpose of PayPal is to allow people to checkout without a user record in our database. This extension has support for being placed in the cart and we should continue supporting this, especially if we want to be listed as official PayPal extension. This feature is mandatory for that. That’s also true for Apple Pay anyway.

mdesantis:
@tvdeyen so the current view implementation is lack of proper PayPal vault flow? I'm confused by that; the current implementation doesn't have a flow where you can skip user creation. It hasn't been implemented yet?

mdesantis:
I'm sorry, I didn't read this: https://github.com/solidusio/solidus_paypal_braintree#express-checkout-from-the-cart

kennyadsl:
@mdesantis can we close this one?

@tvdeyen I can't recall or find where we started deprecating `last_name` in favor of storing everything into `first_name`, can you please point me to any code or documentation about that?

mdesantis:
@kennyadsl if we agree that this isn't the right way to fix #226 we can close it, but I'm unsure; maybe you guys have a stronger opinion about that?

seand7565:
Hey all! Just to weigh in here (a year later): The address-import feature is definitely mandatory as far as PayPal is concerned, and as far as the name issue - firstname and lastname have been deprecated in favor of just `name` here: https://github.com/solidusio/solidus/pull/3584

I think we can safely close this issue - though we should definitely make sure this extension will still work when the deprecation expires 😨

PR #230, opened 2019-06-22
Update italian translations

delphaber:
Updated italian translations. Used `en.yml` as base.

Added `paypal_flow` as a new translation for `en` too.


jacobherrington:
Paging @spaghetticode and @kennyadsl 🇮🇹

PR #238, opened 2019-09-25
Make use of data collector in JS configurable

mamhoff:
Some stores do not want or need to do invasive fraud protection. For
these, it should be possible to turn data collection off without
overwriting partials.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,7 @@ Payment methods can accept preferences either directly entered in admin, or from
         public_key: ENV['BRAINTREE_PUBLIC_KEY'],
         private_key: ENV['BRAINTREE_PRIVATE_KEY'],
         paypal_flow: 'vault', # 'checkout' is accepted too
+        use_data_collector: true # Fingerprint the user's browser when using Paypal
       }
     )
   end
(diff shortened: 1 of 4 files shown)


mamhoff:
What about we go ahead with this for now and create the configuration object once #236 is merged?

tvdeyen:
@mamhoff agree. #236 has now been merged.

kennyadsl:
Hey @mamhoff, I think this needs to be revisited after #236, right?

seand7565:
Hey @mamhoff - would you have time to update this PR? I know it's a year too late, but I'd like to have this merged in (minus the style stuff which was handled in #236) If not just let me know and I'll update this and create a new PR. Thanks!

mamhoff:
@seand7565 I think this does the trick - sorry I took a bit longer :)

seand7565 (commented):
This looks great, thanks! Just one thing - can you add something to the README about this option? Maybe under the PayPal section, something like:
`For fraud prevention, PayPal recommends using a data collector to collect device information, which we've included. You're able to turn off the PayPal data collector on the payment method options if you desire.`

mamhoff:
@seand7565 Added the documentation!

seand7565 (approved):
Looks great, thanks!

kennyadsl:
@mamhoff can you please rebase now that we have specs running against 3.0 as well? Thanks!

kennyadsl (approved):

tvdeyen (approved):
Now that Kount and braintree do not work together anymore this will most likely need to change soon. but for now this is ok.

PR #248, opened 2020-01-31
Upgrade the extension using solidus_dev_support

blocknotes:
Related to "Adopt solidus_dev_support in all extensions" (from solidus_dev_support):
https://github.com/solidusio/solidus_dev_support/issues/10


kennyadsl:
Do we also need to move decorators into the right folders?

blocknotes:
> Do we also need to move decorators into the right folders?

Good point, I make the change.

Issue #244, opened 2019-11-22
Installer breaks during migration


hrdchz:
It appears that #236 breaks the installer? I get an error about `preferences` when running the migrations, which was introduced in those commits.

Error message:

```
=== 20191122221009 AddBraintreeConfigurationToStores: migrating ================
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:

undefined local variable or method `preferences' for #<SolidusPaypalBraintree::Configuration:0x00007f99466f4b18>
Did you mean?  preference_type
/home/asdf/vagrants/s.jumble/db/migrate/20191122221009_add_braintree_configuration_to_stores.solidus_paypal_braintree.rb:4:in `up'
/home/asdf/.rbenv/versions/2.6.2/bin/bundle:23:in `load'
/home/asdf/.rbenv/versions/2.6.2/bin/bundle:23:in `<main>'

Caused by:
NameError: undefined local variable or method `preferences' for #<SolidusPaypalBraintree::Configuration:0x00007f99466f4b18>
Did you mean?  preference_type
/home/asdf/vagrants/s.jumble/db/migrate/20191122221009_add_braintree_configuration_to_stores.solidus_paypal_braintree.rb:4:in `up'
/home/asdf/.rbenv/versions/2.6.2/bin/bundle:23:in `load'
/home/asdf/.rbenv/versions/2.6.2/bin/bundle:23:in `<main>'
Tasks: TOP => db:migrate
(See full trace by running task with --trace)

```

rails 5.2.3 / ruby 2.6.2 / solidus 2.9.3

I was able to install with `gem 'solidus_paypal_braintree', github: 'solidusio/solidus_paypal_braintree', ref: '3e7bb8074'`

luaysuarna:
Any lucky here? I found same problem

luaysuarna:
I skipped this migration
```
# This migration comes from solidus_paypal_braintree (originally 20161125172005)
class AddBraintreeConfigurationToStores < SolidusSupport::Migration[4.2]
  def up
    Spree::Store.all.each(&:create_braintree_configuration)
  end

  def down
    SolidusPaypalBraintree::Configuration.joins(:store).destroy_all
  end
end

```

and finish all migration, then running back tthe command from rails console. All done! :)

waymondo:
Also noticing this is still broken on master

bazfer:
Any plans to roll back the breaking change? Rather have the installation working properly than an extra feature.

[...]

PR #249, opened 2020-02-26
Remove AddBraintreeConfigurationToStores migration

MinasMazar:
because store's braintree configuration doesn't already have
paypal_button_preferences fields and validations here will break migration

Should fix https://github.com/solidusio/solidus_paypal_braintree/issues/244 , while failing tests should be fixed in #253 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ SolidusPaypalBraintree::Gateway.new(
 ```
 
 ### Configure payment types
-Your payment method can accept payments in three ways: through Paypal, through ApplePay, or with credit card details entered directly by the customer. By default all are disabled for all your site's stores.
+Your payment method can accept payments in three ways: through Paypal, through ApplePay, or with credit card details entered directly by the customer. By default all are disabled for all your site's stores. Before proceeding to checkout, ensure you've created a Braintree configuration for your store:
 
 1. Visit /solidus_paypal_braintree/configurations/list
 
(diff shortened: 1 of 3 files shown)


db/migrate/20161125172005_add_braintree_configuration_to_stores.rb
```diff
@@ -1,9 +0,0 @@
-class AddBraintreeConfigurationToStores < SolidusSupport::Migration[4.2]
-  def up
-    Spree::Store.all.each(&:create_braintree_configuration)
-  end
-
-  def down
-    SolidusPaypalBraintree::Configuration.joins(:store).destroy_all
-  end
-end
```
elia: Wouldn't it be better to empty the migration instead of removing it, I mean for existing apps that already have embedded it and might get confused otherwise.

README.md
```diff
@@ -73,7 +73,7 @@ SolidusPaypalBraintree::Gateway.new(
 ```
 
 ### Configure payment types
-Your payment method can accept payments in three ways: through Paypal, through ApplePay, or with credit card details entered directly by the customer. By default all are disabled for all your site's stores.
+Your payment method can accept payments in three ways: through Paypal, through ApplePay, or with credit card details entered directly by the customer. By default all are disabled for all your site's stores. Before proceeed to checkout, ensure you've created a Braintree configuration for your store:
```
seand7565, line 76: `proceeed` -> `proceeding`

seand7565 (changes requested):
Just one small typo in the README to fix, but other than that this looks great. Thanks @MinasMazar !

MinasMazar:
All comments have been addressed (sorry for long delay). cc @seand7565 @elia

PR #250, opened 2020-02-26
Use relative Response class in SolidusPaypalBraintree::Gateway

thefotios:
My team hit an issue where we were using this in a module that has `Response` mixed in already (I believe it was from `ActionView`), causing this to error out with things like `method_not_found` since it was resolving the wrong class.

Given how overloaded the `Response` name is, it's likely other people will hit this as well. By using `require_relative`, it ensures that the correct class is resolved regardless of how and where it's called from.


octave:
👍

octave (approved):
Yes please merge!

thefotios:
Bump. I just merged some new test fixes from upstream. The previous postgres failure seemed unrelated to the PR changes; hopefully this run will be :green_circle:

kennyadsl:
Thanks for the contribution and sorry for the long wait. 

Why adding the `require` should fix the issue? If another `Response` object is there what gives to the Braintree one precedence? I'd have simply prefixed `SolidusPaypalBraintree::` to `Response` in all occurrences of that file but maybe I'm missing something?

PR #252, opened 2020-02-27
Fix Braintree checkout helpers

### LINKED ISSUE #251: Paypal button preference helper not found
Testing on a brand new solidus app, in the payments step I'm running into this error:

```ruby
undefined method `paypal_button_preference' for #<Class:0x00007f2074d2e560>:0x00007f206cf96fa8>
```

Extracted from stacktrace:

```ruby
    25  locale: '<%= paypal_button_preference(:paypal_button_locale, store: current_store) %>',
```

MinasMazar:
This should fix #251 

Using `Spree::CheckoutController.helper` and `Spree::OrdersController.helper` in `config/initializers/braintree.rb` to include helper modules does not work: at `checkout/payement` step an exception is raised

```
ActionView::Template::Error (undefined method `paypal_button_preference`)
```

This commit move `paypal_button_preferences` (introduced in #236) in the already present `SolidusPaypalBraintree::BraintreeCheckoutHelper` module.

The same is fix is done for the [`BraintreeAdminHelper`](https://github.com/solidusio/solidus_paypal_braintree/blob/master/app/helpers/braintree_admin_helper.rb) module, which was affected by the same bug.


config/initializers/braintree.rb
```diff
@@ -1,8 +1,3 @@
 if SolidusSupport.backend_available?
   Spree::Admin::PaymentsController.helper :braintree_admin
 end
-
-if SolidusSupport.frontend_available?
-  Spree::CheckoutController.helper :braintree_checkout
-  Spree::OrdersController.helper :braintree_checkout
-end
```
elia: Can you deal with this in a separate PR? 
Maybe check if the backend helper above can also be removed. 🙏

seand7565:
@MinasMazar Would you have some time to finish this PR? I'm starting to work on this repo a bit, and this is definitely a big issue! If not, just let me know - I can take over the excellent work done here. Thanks!

PR #255, opened 2020-04-27
Bump solidus_support to ~> 0.5

peterberkenbosch:


seand7565:
Looks like #261 handled this - thanks for the PR @peterberkenbosch , sorry it took so long to address it!

PR #260, opened 2020-07-28
Fix LoadError on failed payments

mamhoff:
We're getting the following error in production:

vendor/bundle/ruby/2.6.0/bundler/gems/solidus_paypal_braintree-091a6f94203a/app/models/solidus_paypal_braintree/gateway.rb:210:in `create_profile': uninitialized constant SolidusPaypalBraintree::Spree::Core (NameError)

I believe adding the root namespace `::` would fix those.


mamhoff:
This does not, actually, fix the problem.

PR #263, opened 2020-10-08
Fix value conversion method

seand7565:
In Solidus, this method has changed from having two parameters to
having three. This updates the method here to match, so we avoid
any too_many_parameter errors. Most noticably, this was affecting
our test suite.

See: https://github.com/solidusio/solidus/commit/e05da68179b45a68d762dfcc7c1ba6618d00d234

### DIFF
--- a/app/models/solidus_paypal_braintree/gateway.rb
+++ b/app/models/solidus_paypal_braintree/gateway.rb
@@ -292,11 +292,15 @@ module SolidusPaypalBraintree
       JSON.parse(preference_string.gsub("=>", ":"))
     end
 
-    def convert_preference_value(value, type)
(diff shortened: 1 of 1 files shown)


seand7565:
Ah - didn't consider that it supers, so on older versions of Solidus this will still fail. Stand by, will figure something out...

seand7565:
Due to the awkwardness of the change from 2 to 3 parameters, and needing to support multiple Solidus versions, we're using a method here that checks how many parameters `super` has, and acting accordingly.

Should be good for a review now!

kennyadsl (approved):
We should probably add support to convert a string to a hash in core directly, but that's another story. Works for me!

PR #264, opened 2020-10-08
Fix 3DS testing

seand7565:
For whatever reason, the listed Braintree cards that are _supposed_
to trigger the fail/pass conditions do not anymore. Instead, using the
cards listed in Cardinal documentation does trigger those conditions.

The structure of the 3ds window is also slightly different, with the
input being nested in two iframes now.


seand7565:
Well... It works locally!  🤷 Back to the drawing board.

kennyadsl (approved):
Thanks Sean!

PR #267, opened 2020-10-14
Update to New PayPal SDK

seand7565:
This switches from the braintree SDK to the new PayPal SDK, which
allows for displaying credit messaging/buttons.

This update adds two new config options:
* Layout, accepting either `horizontal` or `vertical`, changing
the layout of the buttons
* Messaging, a boolean, that controls whether the messaging
component is loaded in.

This update also removes the `size` config option, as it no longer
does anything. PayPal uses the responsive size for everything now,
and expects you to resize your elements locally to whatever size
you want.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,12 @@ It will only be displayed if the `SolidusPaypalBraintree::Gateway` payment
 method is configured to display on the frontend and PayPal is enabled in the
 store's configuration.
 
+You can find button configuration options in 
+`/solidus_paypal_braintree/configurations/list` if you want to change the color,
+shape, layout, and a few other options. Keep in mind that the `paypal_button_tagline`
+does not work when the `paypal_button_layout` is set to `vertical`, and will be
+ignored in that case.
+
 The checkout view
 [initializes the PayPal button](/lib/views/frontend/spree/checkout/payment/_paypal_braintree.html.erb)
 using the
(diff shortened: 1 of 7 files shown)


kennyadsl:
Needs a rebase now that I merged the Rubocop one, @seand7565 .

kennyadsl (approved):
Just one thing: are some of these change breaking for who already uses the extensions? If yes we should either keep the old code working or release a major.

seand7565:
> Just one thing: are some of these change breaking for who already uses the extensions? If yes we should either keep the old code working or release a major.

I could see two cases where this would be considered breaking:
If you've overridden the PayPal button create to use your own methods, those methods will need to be updated from the old `payment` and `on_authorize` returns to the new `createOrder/createBillingAgreement` and `onApprove` methods

-and-

The `size` attribute no longer does anything for PayPal, meaning that some stores will have to do some CSS work to get it looking normal again. Come to think of it, the added messaging component might also mean looking at CSS again depending on the store.

The changes that make the above possible all come from updating to the new PayPal SDK, so keeping the old code around wouldn't help I think. IMO this change warrants a new release. WDYT?

kennyadsl:
I agree, and I'm seeing that we have no major for this extension. We can still do a minor bump to 0.5.0 hoping it's enough to indicate that there's something breaking.

PR #271, opened 2020-10-22
Update transaction_address to support name attribute

### LINKED ISSUE #226: On PayPal payment user shipping address gets set with uncorrect firstname and lastname values if they contain a space

seand7565:
Solidus is moving away from using first_name and last_name, in favor of
just "name". This updates the paypal JS to send a name to
transaction_addess, rather than splitting it on the spot.
transaction_address is responsible for either:

* Splitting the name, if the user is on Solidus 2.10 or below
-or-
* Passing the name directly to the address, if the user is on Solidus
2.11 or higher - where the name parameter is supported.

Also deprecates using first_name and last_name in transaction_address,
in case someone has overridden `paypal_button`

Fixes #226

### DIFF
--- a/app/models/solidus_paypal_braintree/transaction_address.rb
+++ b/app/models/solidus_paypal_braintree/transaction_address.rb
@@ -8,11 +8,11 @@ module SolidusPaypalBraintree
     include ActiveModel::Validations::Callbacks
     include SolidusPaypalBraintree::CountryMapper
 
-    attr_accessor :country_code, :last_name, :first_name,
-      :city, :zip, :state_code, :address_line_1, :address_line_2
+    attr_accessor :country_code, :name, :city, :zip, :state_code,
+      :address_line_1, :address_line_2, :first_name, :last_name
 
-    validates :first_name, :last_name, :address_line_1, :city, :zip,
-      :country_code, presence: true
+    validates :address_line_1, :city, :zip, :country_code, presence: true
+    validates :name, presence: true, unless: ->(address){ address.first_name.present? }
 
     before_validation do
       self.country_code = country_code.presence || "us"
@@ -44,8 +44,6 @@ module SolidusPaypalBraintree
 
     def to_spree_address
       address = ::Spree::Address.new(
-        first_name: first_name,
-        last_name: last_name,
         city: city,
         country: spree_country,
         address1: address_line_1,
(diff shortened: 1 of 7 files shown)


app/models/solidus_paypal_braintree/transaction_address.rb
```diff
@@ -44,15 +44,25 @@ def spree_state
...
       )
 
+      if !first_name.nil?
+        ::Spree::Deprecation.warn("first_name and last_name are deprecated. Use name instead.", caller)
+        address.first_name = first_name
+        address.last_name = last_name || "(left blank)"
+      elsif address.respond_to? :name
+        address.name = name
+      else
+        first_name, last_name = split_name(name)
```
kennyadsl, line 61: Don't we have something in Solidus Core that does the same thing? I recall we have a class for that

seand7565: In Solidus 2.11, there's a `Spree::Address::Name` virtual attribute which handles splitting, it's what line 58-59 handle. I'm not aware of any class that handles name splitting before 2.11 - at least, I took a quick look just now and couldn't find one!

kennyadsl: I think we should use that class if it's defined. What do you think?

seand7565: `address.name` creates a new instance of `Spree::Address::Name` and sets the first and last names accordingly: https://github.com/solidusio/solidus/blob/44da6ff1061ac8dfd339f02228bbd921b4a4c006/core/app/models/spree/address.rb#L204-L210 - so above, if that class is available, we're using it as the name splitter.

kennyadsl: Oh, you are 100% right, thanks Sean!

kennyadsl (approved):

aldesantis (approved):

Issue #121, opened 2017-08-30
Allow passing of styles and placeholder text into braintree hosted fields


joeljackson:
As per for example

https://codepen.io/braintree/pen/mPgdPN

kennyadsl:
We can probably do something similar to this: https://github.com/solidusio/solidus_paypal_braintree/pull/238

PR #272, opened 2020-10-27
Add ability to style braintree credit card fields

seand7565:
Using the `credit_card_fields_style` preference, you're able to pass
style attributes to the hosted fields. This allows you to pass any
arbitrary CSS to the Braintree fields, such as font size or color.

Handles the style portion of #121


seand7565:
Converting to draft - I need to readdress this PR a bit.

PR #276, opened 2020-10-29
Allow user to specify style and placeholder_text of hosted_fields

### LINKED ISSUE #121: Allow passing of styles and placeholder text into braintree hosted fields

seand7565:
Adds the `credit_card_fields_style` and `placeholder_text` preferences to the payment_method, allowing the user full control over the look & feel of their hosted fields. 

closes #121

### DIFF
--- a/lib/views/backend/spree/admin/payments/source_forms/_paypal_braintree.html.erb
+++ b/lib/views/backend/spree/admin/payments/source_forms/_paypal_braintree.html.erb
@@ -11,6 +11,6 @@
   </div>
 
   <div id="card_form<%= id %>" data-hook style="display: none;">
-    <%= render partial: "spree/shared/braintree_hosted_fields", locals: { id: id } %>
+    <%= render partial: "spree/shared/braintree_hosted_fields", locals: { payment_method: payment_method } %>
   </div>
 </fieldset>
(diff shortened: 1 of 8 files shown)


lib/views/backend/spree/admin/payments/source_forms/_paypal_braintree.html.erb
```diff
@@ -11,6 +11,6 @@
   </div>
 
   <div id="card_form<%= id %>" data-hook style="display: none;">
-    <%= render partial: "spree/shared/braintree_hosted_fields", locals: { id: id } %>
+    <%= render partial: "spree/shared/braintree_hosted_fields", locals: { payment_method: payment_method } %>
```
kennyadsl, line 14: Would be great if we can keep the partial working for who is rendering it in their application in other places. What if we keep accepting the old `id` key as well and deprecate it inside the view?

seand7565: Yeah, great idea! How does the new version look?

kennyadsl (approved):

PR #278, opened 2020-11-02
Disable PayPal messaging when vault flow is enabled

seand7565:
PayPal doesn't want you to use the messaging component when vault flow
is being used. This disables the default messaging component when vault
flow is on, and defaults the component to disabled (since the extensions
defaults to vault flow: true)

### DIFF
--- a/app/models/solidus_paypal_braintree/configuration.rb
+++ b/app/models/solidus_paypal_braintree/configuration.rb
@@ -8,7 +8,7 @@ module SolidusPaypalBraintree
       label: { availables: %w[checkout credit pay buynow paypal installment], default: 'checkout' },
       tagline: { availables: %w[true false], default: 'false' },
       layout: { availables: %w[horizontal vertical], default: 'horizontal' },
-      messaging: { availables: %w[true false], default: 'true' }
+      messaging: { availables: %w[true false], default: 'false' }
     }.freeze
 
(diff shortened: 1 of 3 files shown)


app/models/solidus_paypal_braintree/configuration.rb
```diff
@@ -8,7 +8,7 @@ class Configuration < ::Spree::Base
       label: { availables: %w[checkout credit pay buynow paypal installment], default: 'checkout' },
       tagline: { availables: %w[true false], default: 'false' },
       layout: { availables: %w[horizontal vertical], default: 'horizontal' },
-      messaging: { availables: %w[true false], default: 'true' }
+      messaging: { availables: %w[true false], default: 'false' }
```
kennyadsl, line 11: If it's disabled by default, should we explain in the readme how to enable messaging?

seand7565: I added this to the `paypal financing messaging` section:
```
To enable this option, you'll need to use the `checkout` flow, and set the `paypal button messaging` option to `true` in your Braintree configuration.
```
Does that work?

kennyadsl: Oh right, sorry!

kennyadsl (approved):

PR #282, opened 2020-11-09
Use messaging partial on payment buttons

seand7565:
Uses the existing messaging partial on the cart & checkout payment
buttons, and passes the total & placement along. Also adds a
`skip_initialization` option to the partial, because we don't want to
initialize the messaging component twice, and PayPal _really_ doesn't
like loading in their SDK twice, even if for two different components.

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,6 +197,21 @@ You can also include this view partial to implement this messaging component any
 render "spree/shared/paypal_messaging, options: {total: @product.price, placement: "product", currency: 'USD'}"
 ```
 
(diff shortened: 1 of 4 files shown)


kennyadsl (commented):
Is there any occurrence in the code when we are using `paypal_messaging` partial initializing the head scripts? What about always skipping it and make it clear in the README that it has to be done manually?

seand7565:
No, we're not using it anywhere without PayPal already being initialized. I think that's a good idea - one sec, I'll update this.

PR #286, opened 2021-04-02
Update to Solidus 3.0

MinasMazar:
#### Let the extension work with Solidus 3.0.

Adjust code related to `Spree::Address` to fit the new [_combined first and last name_ behaviour](https://github.com/solidusio/solidus/pull/3820); it also update some dependency and lint rules.

### DIFF
--- a/app/models/solidus_paypal_braintree/transaction_address.rb
+++ b/app/models/solidus_paypal_braintree/transaction_address.rb
@@ -21,15 +21,6 @@ module SolidusPaypalBraintree
     validates :spree_country, presence: true
     validates :state_code, :spree_state, presence: true, if: :should_match_state_model?
 
-    def self.split_name(name)
-      if defined?(Spree::Address::Name)
-        address_name = Spree::Address::Name.new(name)
-        [address_name.first_name, address_name.last_name]
-      else
-        name.strip.split(' ', 2)
-      end
-    end
-
     def initialize(attributes = {})
       country_name = attributes.delete(:country_name) || ""
       if attributes[:country_code].blank?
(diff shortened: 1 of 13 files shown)


app/models/solidus_paypal_braintree/transaction_address.rb
```diff
@@ -51,16 +51,24 @@ def to_spree_address
...
 
-      if !first_name.nil?
-        ::Spree::Deprecation.warn("first_name and last_name are deprecated. Use name instead.", caller)
-        address.first_name = first_name
-        address.last_name = last_name || "(left blank)"
-      elsif address.respond_to? :name
-        address.name = name
+      if SolidusSupport.combined_first_and_last_name_in_address?
+        address.name = begin
+          if first_name.nil?
```
kennyadsl, line 56: Why this check? In Solidus 2.11 we'll have both `first_name` and `name` always synced and this is probably always false. In Solidus 3, there's no reason to check `first_name`, which is no more used.

kennyadsl (commented):
Left a question, thanks!

app/models/solidus_paypal_braintree/transaction_address.rb
MinasMazar: Here `first_name` is an instance attribute of `SolidusPaypalBraintree::TransactionAddress` (together with `last_name` and `name`). If I recall correctly these values are returned by Braintree itself and maybe they can be valorized differently according to the API version (that's why we need the check on its presence).

kennyadsl: Ah makes sense!

kennyadsl (approved):

PR #287, opened 2021-04-08
Use State Abbreviation in Checkout Helper

mamhoff:
Braintree states on their website that "any string up to 50 characters
is valid". This is not true, according to

https://stackoverflow.com/questions/65701058/sudden-failure-requests-on-braintree-sandbox-api-billing-state-format-is-inval

So apparently Braintree ACTUALLY wants a two-letter abbreviation.
We've tested this manually...


kennyadsl:
We addressed this in #286 , thanks anyway.

PR #288, opened 2021-04-08
Include Persistable Module in SolidusPayPalBraintree::Configuration

mamhoff:
This stops us from getting a deprecation warning on Solidus 2.11, and from outright failing to build on 3.0.


kennyadsl:
We addressed this in https://github.com/solidusio/solidus_paypal_braintree/pull/286/commits/128ddddd26fd6146f82ad785f1eb48404fbc5b7a. Do you think it's a viable solution as well?

mamhoff:
Sure, it's good enough!

PR #289, opened 2021-04-09
Make extension compliant to solidus_dev_support

MinasMazar:


kennyadsl:
@MinasMazar this is still needed after merging #286 ?

RyanofWoods:
This could great to move forward @kennyadsl & @MinasMazar, not having a sandbox on this extension makes developing quite a pain 🙁

MinasMazar:
hey sorry, this work has been stale for a long time :\  @kennyadsl @RyanofWoods if I recall correctly #286 was only addressing some changes required for Solidus 3.0 concerning the `Spree::Address` model. Here I was adding the needed executables generated by `solidus_dev_support`, making the extension fully compliant with dev support guidelines, so I think we still need it to be finished. I'll try to revamp this PR in the next few days and then ask you for a double-check.

lib/solidus_paypal_braintree/testing_support/factories.rb
```diff
@@ -0,0 +1,4 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
```
RyanofWoods, line 3: Should we move the existing factories into here @MinasMazar?
https://github.com/solidusio/solidus_paypal_braintree/blob/master/lib/solidus_paypal_braintree/factories.rb

MinasMazar: You're right! Updated [here](https://github.com/solidusio/solidus_paypal_braintree/pull/289/commits/276386929c928efa68e1b059a6df184f13874ba0).

kennyadsl:
Thanks @MinasMazar, if you can squash the fixup I think we are there!

kennyadsl (approved):

RyanofWoods:
Thanks @MinasMazar! 🙌

PR #291, opened 2021-05-21
Set sandbox mode if environment is sandbox

alexblackie:
Currently if environment is passed in the options, it is set to a
prototype field and then promptly ignored and never used. This
configures the paypal client instance to use sandbox mode (through
loading with `client-id=sb` [as per the docs][1]) if the caller has
specified the environment to be "sandbox."

[1]: https://developer.paypal.com/docs/checkout/reference/customize-sdk/#client-id

### DIFF
--- a/app/assets/javascripts/solidus_paypal_braintree/paypal_button.js
+++ b/app/assets/javascripts/solidus_paypal_braintree/paypal_button.js
@@ -41,6 +41,7 @@ SolidusPaypalBraintree.PaypalButton.prototype.initializeCallback = function() {
   this._paymentMethodId = this._client.paymentMethodId;
 
   this._client.getPaypalInstance().loadPayPalSDK({
+    "client-id": this._environment === "sandbox" ? "sb" : null,
     currency: this._paypalOptions.currency,
     commit: true,
     vault: this._paypalOptions.flow == "vault",


swively (approved):

app/assets/javascripts/solidus_paypal_braintree/paypal_button.js
```diff
@@ -41,6 +41,7 @@ SolidusPaypalBraintree.PaypalButton.prototype.initializeCallback = function() {
   this._paymentMethodId = this._client.paymentMethodId;
 
   this._client.getPaypalInstance().loadPayPalSDK({
+    "client-id": this._environment === "sandbox" ? "sb" : null,
```
kennyadsl, line 44: What about configuring this at `_paypalOptions` level instead? I see that a lot of other configurations are done this way.

kennyadsl (commented):
Thanks, just left a question.

app/assets/javascripts/solidus_paypal_braintree/paypal_button.js
alexblackie: `this._environment` does come from the `paypalOptions`, it just was never being used here.

kennyadsl (approved):

PR #299, opened 2021-06-23
Update to Braintree 3.4.0

pelargir:
This updates the Braintree gem from v2.65 to v3.4. Among other benefits, this gives better insight into payment and refund declines by adding a processor response code in the API response object.

See https://www.braintreepayments.com/blog/refund-authorizations-api-updates-and-sandbox-testing/


mamhoff:
This would be fantastic to get at least into the main branch if not released. Pretty please?

kennyadsl (approved):

kennyadsl:
Thanks @pelargir

PR #306, opened 2021-11-24
Require MFA for RubyGems actions

RyanofWoods:
This setting tells RubyGems that MFA is required for accounts to be able
perform any of these privileged operations:
- gem push
- gem yank
- gem owner --add/remove
- adding or removing owners using gem ownership page

This helps make the gem more secure, as users can be more confident
that gem updates were pushed by maintainers.

More information:
https://guides.rubygems.org/mfa-requirement-opt-in/
https://docs.rubocop.org/rubocop/cops_gemspec.html#gemspecrequiremfa


DanielePalombo (approved):
👍

PR #313, opened 2022-01-17
Fix rubocop warnings

mamhoff:
With recent updates of `rubocop-rspec` and `rubocop-rails`, we got a few new warnings that this PR addresses - some through following the cop's recommendations, some through disabling them.


tvdeyen (approved):
Thank you!

PR #314, opened 2022-01-17
Lower ActiveRecord::Migration version in add_paypal_funding_source

mamhoff:
The migration does not need any new features of Rails 6.1, so we can
safely use 5.0 and be backwards-compatible.


RyanofWoods:
Sorry about that @mamhoff! Thank you for addressing and making a PR for the fix. I will remember this in the future 🚀

mamhoff:
Don't worry @RyanofWoods , nothing to be sorry about :)

tvdeyen (approved):
Thanks.

kennyadsl (approved):
Thanks @mamhoff!

PR #316, opened 2022-03-04
Do not safeguard including of Spree::Preferences::Persistable

tvdeyen:
The safeguard prevents this module to be included because the
class responds to `preference` (it is included in `Spree::Base`).

My guess is that this safeguard was included to avoid including
`Spree::Preferences::Preferable` twice (it is included by `Spree::Preferences::Persistable`)

But Ruby will take care of not including the same module twice.

This will not effect user of Solidus >= 3.0, but helps to get rid of a (rather large) deprecation warning in Solidus 2.11

### DIFF
--- a/app/models/solidus_paypal_braintree/configuration.rb
+++ b/app/models/solidus_paypal_braintree/configuration.rb
@@ -11,9 +11,7 @@ module SolidusPaypalBraintree
       messaging: { availables: %w[true false], default: 'false' }
     }.freeze
 
-    unless respond_to?(:preference)
-      include ::Spree::Preferences::Persistable
-    end
+    include ::Spree::Preferences::Persistable
 
     belongs_to :store, class_name: 'Spree::Store', optional: false
 


mamhoff (commented):
Not so sure about Ruby not including things twice: 
```rb
[9] pry(main)> module Foo
[9] pry(main)*   def self.included(klass)
[9] pry(main)*     puts "INCLUDED"
[9] pry(main)*   end  
[9] pry(main)* end  
=> :included
[10] pry(main)> include Foo
INCLUDED
=> Object
[11] pry(main)> include Foo
INCLUDED
=> Object
[12] pry(main)> 
```

tvdeyen:
> Not so sure about Ruby not including things twice:
> 
> ```ruby
> [9] pry(main)> module Foo
> [9] pry(main)*   def self.included(klass)
> [9] pry(main)*     puts "INCLUDED"
> [9] pry(main)*   end  
> [9] pry(main)* end  
> => :included
> [10] pry(main)> include Foo
> INCLUDED
> => Object
> [11] pry(main)> include Foo
> INCLUDED
> => Object
> [12] pry(main)> 
> ```

Ok, maybe I did not phrased it correctly. Ruby will not append the same methods to the same class twice. Also this is a `ActiveSupport::Concern` and this will take care of other potential side effects of the `include`.

app/models/solidus_paypal_braintree/configuration.rb
```diff
@@ -11,9 +11,7 @@ class Configuration < ::Spree::Base
       messaging: { availables: %w[true false], default: 'false' }
     }.freeze
 
-    unless respond_to?(:preference)
```
kennyadsl, line 14: @MinasMazar any chance you recall why we added this check [here](https://github.com/solidusio/solidus_paypal_braintree/commit/128ddddd26fd6146f82ad785f1eb48404fbc5b7a)? 🙏

kennyadsl (approved):
If we don't why this is here I think we can remove it.

app/models/solidus_paypal_braintree/configuration.rb
MinasMazar: @kennyadsl I'm sorry, but I'm not aware right now. I think we can remove it.

elia (approved):

PR #318, opened 2022-04-29
Fix exception when other payment methods active

embold-tyler:
When other payment methods are active, specifically solidus_affirm, these methods are not available and cause an exception.

### DIFF
--- a/lib/views/frontend/solidus_paypal_braintree/payments/_payment.html.erb
+++ b/lib/views/frontend/solidus_paypal_braintree/payments/_payment.html.erb
@@ -1,12 +1,12 @@
 <br />
-<%= payment.source.display_payment_type %>
+<%= payment.source.try(:display_payment_type) %>
 
-<% if payment.source.paypal? %>
+<% if payment.source.try(:paypal?) %>
   <% if payment.source.respond_to?(:paypal_funding_source) && payment.source.paypal_funding_source.present? %>
     <br />
     <%= t('spree.paypal_funding', funding: payment.source.display_paypal_funding_source) %>
   <% end %>
-<% elsif payment.source.venmo? %>
+<% elsif payment.source.try(:venmo?) %>
   <br />
   <%= payment.source.source_description %>
 <% end %>


kennyadsl:
@embold-tyler thanks for the PR. Can you help me better understand this issue, please? That partial should be rendered for payment associated with this PaypalBraintree payment method only. In your case it seems like it's used for an affirm payment method, right?

embold-tyler:
No problem. I noticed that the deface is going onto the payments view, so it hits each payment method and it seems the methods in those conditionals its using are not available on the payments sources generated by affirm for whatever reason ("display_payment_type", "paypal?", and "venmo?") so was causing the app to throw a no method error. Using .try allowed the conditional a graceful fail while still rendering those portions of the defaced partial only when needed on the PaypalBraintree method.

kennyadsl:
Oh, which deface exactly? I can't see any deface that could interact with other payment partials in this extension. Maybe it's coming from somewhere else?

embold-tyler:
This one: https://github.com/solidusio/solidus_paypal_braintree/blob/master/app/overrides/spree/payments/payment/add_paypal_funding_source_to_payment.rb

kennyadsl:
But it is explicitely applied to the Paypal partial only, [see line 8](https://github.com/solidusio/solidus_paypal_braintree/blob/42e3a0a7623989d288ce81605c8e403b5161ae49/app/overrides/spree/payments/payment/add_paypal_funding_source_to_payment.rb#L8).

kennyadsl:
I think you are right, I was reading that deface override wrongly. The partial I linked is what is being injected but it is going in the generic payment partial coming from Solidus core. I think your fix addresses the issue but I'd like to be sure to render those things only for payments coming from this extension.

RyanofWoods:
I tried this locally, and this change looks good to me @kennyadsl.

Thank you for the fix @embold-tyler

RyanofWoods:
Can we release this patch @kennyadsl and what about yanking 1.1.0 and 1.1.1 because 1.1.0 was a breaking change.

kennyadsl:
@RyanofWoods I'll release it now. I'm against yanking those versions though. Yanking in my experience causes more problems than it solves. Having 1.1.2 would be enough to prevent people from updating to 1.1.0 or 1.1.1. If they are locked to those versions, they are probably not experiencing this bug.

kennyadsl:
@RyanofWoods @embold-tyler 1.1.2 released, let me know how it goes. Thanks again!

RyanofWoods:
Understood, thanks @kennyadsl 😊

PR #321, opened 2022-06-16
Make gem compatible with Rails 7

gsmendoza:
## Main changes

* Fix `register_solidus_paypal_braintree_gateway` initializer for Rails 7. See https://guides.rubyonrails.org/autoloading_and_reloading_constants.html.
* Fix transaction address validation error handling

### DIFF
--- a/lib/solidus_paypal_braintree/engine.rb
+++ b/lib/solidus_paypal_braintree/engine.rb
@@ -21,8 +21,8 @@ module SolidusPaypalBraintree
     initializer "register_solidus_paypal_braintree_gateway", after: "spree.register.payment_methods" do |app|
       config.to_prepare do
         app.config.spree.payment_methods << SolidusPaypalBraintree::Gateway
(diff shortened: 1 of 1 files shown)


lib/solidus_paypal_braintree/engine.rb
```diff
@@ -19,9 +19,11 @@ class Engine < Rails::Engine
     end
 
     initializer "register_solidus_paypal_braintree_gateway", after: "spree.register.payment_methods" do |app|
-      app.config.spree.payment_methods << SolidusPaypalBraintree::Gateway
-      SolidusPaypalBraintree::Gateway.allowed_admin_form_preference_types << :preference_select
-      Spree::PermittedAttributes.source_attributes.concat [:nonce, :payment_type, :paypal_funding_source]
+      config.to_prepare do
+        app.config.spree.payment_methods << SolidusPaypalBraintree::Gateway
```
gsmendoza, line 23: This `append` call is already idempotent since `app.config.spree.payment_methods` is a Set object

kennyadsl (approved):
Thanks George!

PR #323, opened 2022-07-05
Fix specs to stub spree_current_user

gsmendoza:
Solidus recently migrated all migrated all uses of `try_spree_current_user` to `spree_current_user`. See
https://github.com/solidusio/solidus/commit/201d0e86b2bf97356735e1b90eea0e8df1402671.


kennyadsl (approved):
Thanks @gsmendoza

PR #325, opened 2022-08-21
Fix broken factory

johnpitchko:
Factory uses static assignment, which was deprecated in FactoryBot v5,
so instantiating the factory raises an error for any project using a
recent release of FactoryBot.

Replaced static assignment with dynamic assignment to resolve this
issue.


kennyadsl (approved):
Thanks @johnpitchko!

PR #326, opened 2022-10-10
Drop compatibility with Solidus < 2.4

mamhoff:
This drops support for Solidus version <= 2.3, and stops a deprecation warning from happening when running the Gem under current Solidus versions.


mamhoff:
References https://github.com/solidusio/solidus_support/pull/62

spaghetticode (approved):

PR #328, opened 2022-10-28
Fix broken specs

RyanofWoods:
Having a broken suite is blocking merging PRs as well as giving the
repository a bad reputation, even though the repositories' logic is
not broken.

---

I would be desirable to have different PRs for these commits, however,
unfortunately, they cannot be separated, as each commit fixes a spec,
and a green suite is needed to merge.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -3,7 +3,6 @@ require:
 
 AllCops:
   NewCops: disable
-  TargetRubyVersion: 2.6
 
 Layout/FirstArgumentIndentation:
   EnforcedStyle: consistent
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -2,7 +2,7 @@ require:
   - solidus_dev_support/rubocop
 
 AllCops:
-  NewCops: enable
+  NewCops: disable
   TargetRubyVersion: 2.6
 
 Layout/FirstArgumentIndentation:
--- a/spec/features/frontend/braintree_credit_card_checkout_spec.rb
+++ b/spec/features/frontend/braintree_credit_card_checkout_spec.rb
@@ -7,7 +7,7 @@ shared_context "with frontend checkout setup" do
   let(:three_d_secure_enabled) { false }
   let(:venmo_enabled) { false }
   let(:card_number) { "4111111111111111" }
-  let(:card_expiration) { "01/#{Time.now.utc.year + 2}" }
+  let(:card_expiration) { "01/#{Time.now.utc.year + 3}" }
 
   before do
     braintree.save!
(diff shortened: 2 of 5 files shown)


RyanofWoods:
For https://github.com/solidusio/solidus_paypal_braintree/pull/328/commits/f68d31430dd6a3585b45df02d994f81f571f1e6f I took inspiration for what you did in the PCP extension @elia. Let me know what you think
https://github.com/solidusio/solidus_paypal_commerce_platform/commit/fc16fe7575cbbe6ae00b7a81aaff117adab8772f

elia (approved):
🙌

solidus_paypal_braintree.gemspec
```diff
@@ -14,7 +14,7 @@ Gem::Specification.new do |s|
   s.email     = 'braintree+gemfile@stembolt.com'
   s.homepage  = 'https://github.com/solidusio/solidus_paypal_braintree'
 
-  s.required_ruby_version = '>= 2.5'
+  s.required_ruby_version = '>= 2.6'
```
kennyadsl, line 17: Is this strictly required? I think this might require us to release a major version because we are breaking compatibility with people using Ruby 2.5.

kennyadsl (commented):
Thanks! Left a question about Ruby bump, but overall it's great! ❤️

solidus_paypal_braintree.gemspec
RyanofWoods: I understand. The reason for the change was because the value doesn't match the rubocop version:
https://github.com/solidusio/solidus_paypal_braintree/blob/a27da1b7d5dc9ea52ef75cec3809dbff6d45b66d/.rubocop.yml#L6

and note, 2.6 is no longer supported by Rubocop. So if we want to avoid a major just to satisfy Rubocop, we'll need to disable rubocop for this line.

What do you prefer @kennyadsl?

kennyadsl: I'd say to do the same that we do on other extensions. For example, I don't see the target ruby version specified [here](https://github.com/solidusio/solidus_auth_devise/blob/4a4f9f213d45b77cead29f45bf5bfd0d2021ab46/.rubocop.yml) or [here](https://github.com/solidusio/solidus_stripe/blob/master/.rubocop.yml). I'd go with that if possible.

RyanofWoods:
Hey @elia 👋 Got any ideas on why the `lint-code` job is now failing?

The suite passed originally, but then after I removed the `TargetRubyVersion` from `.rubocop.yml`, it fails.

I don't understand how it can handle a version of `nil` for Rails on some jobs, but not on the `lint-code` one when generating the Gemfile lock 🤔 
https://app.circleci.com/pipelines/github/solidusio/solidus_paypal_braintree/422/workflows/56d0ecc3-1b70-417e-9e89-ee3ce5dcbc30

kennyadsl:
@RyanofWoods please, rebase now 🙂

kennyadsl (approved):

kennyadsl:
@RyanofWoods 

> Hey @elia 👋 Got any ideas on why the lint-code job is now failing?

I see it's passing now!

elia:
Awesome 🙌

PR #329, opened 2022-11-18
Fix sandbox generator

RyanofWoods:
## Summary
Before these changes, the sandbox generator was not working correctly. It would fail to seed data due to the authentication gem not being loaded before the extension. Also, the script was not defaulting to `master` for Solidus.

## Checklist

Check out our [PR guidelines](https://github.com/solidusio/.github/blob/master/CONTRIBUTING.md#pull-request-guidelines) for more details.

The following are mandatory for all PRs:

- [x] I have written a thorough PR description.
- [x] I have kept my commits small and atomic.
- [x] I have used clear, explanatory commit messages.

The following are not always needed (~cross them out~ if they are not):

~~- [ ] I have added automated tests to cover my changes.~~
~~- [ ] I have attached screenshots to demo visual changes.~~
~~- [ ] I have opened a PR to update the [guides](https://github.com/solidusio/edgeguides).~~
~~- [ ] I have updated the README to account for my changes.~~

Screenshot 2022-11-18 at 16 12 03

### DIFF
--- a/bin/sandbox
+++ b/bin/sandbox
@@ -90,7 +90,7 @@ unbundled bundle exec rails generate solidus:install \
   --auto-accept \
   --user_class=Spree::User \
   --enforce_available_locales=true \
-  --with-authentication=true \
+  --authentication="devise" \
   --payment-method=none \
   --frontend=${SOLIDUS_FRONTEND} \
   $@
--- a/bin/sandbox
+++ b/bin/sandbox
@@ -70,6 +70,8 @@ gem 'solidus', github: 'solidusio/solidus', branch: '$SOLIDUS_BRANCH'
 gem 'rails-i18n'
 gem 'solidus_i18n'
 
+# This Braintree extension contains a user decorator, so the user class must be loaded first.
+gem "solidus_auth_devise", "~> 2.5"
 gem '$extension_name', path: '..'
 
 group :test, :development do
@@ -83,6 +85,7 @@ unbundled bundle install --gemfile Gemfile
 
 unbundled bundle exec rake db:drop db:create
 
+# Still request "devise". Solidus will skip installing it again but will include its seeds.
 unbundled bundle exec rails generate solidus:install \
   --auto-accept \
   --user_class=Spree::User \
--- a/bin/sandbox
+++ b/bin/sandbox
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 set -e
-if [ ! -z $DEBUG ]
+if [ -n "$DEBUG" ]
 then
   set -x
 fi
@@ -28,16 +28,14 @@ sqlite3|sqlite)
 esac
 echo "~~> Using $RAILSDB as the database engine"
 
-if [ -n $SOLIDUS_BRANCH ]
+if [ -z "$SOLIDUS_BRANCH" ]
 then
-  BRANCH=$SOLIDUS_BRANCH
-else
   echo "~~> Use 'export SOLIDUS_BRANCH=[master|v3.2|...]' to control the Solidus branch"
-  BRANCH="master"
+  SOLIDUS_BRANCH="master"
 fi
-echo "~~> Using branch $BRANCH of solidus"
+echo "~~> Using branch $SOLIDUS_BRANCH of solidus"
 
-if [ -z $SOLIDUS_FRONTEND ]
+if [ -z "$SOLIDUS_FRONTEND" ]
 then
   echo "~~> Use 'export SOLIDUS_FRONTEND=[solidus_frontend|solidus_starter_frontend]' to control the Solidus frontend"
   SOLIDUS_FRONTEND="solidus_frontend"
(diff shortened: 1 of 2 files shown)


bin/sandbox
```diff
@@ -85,11 +84,12 @@ unbundled bundle install --gemfile Gemfile
 
 unbundled bundle exec rake db:drop db:create
 
+# Still request "devise". Solidus will skip installing it but include the seeds
```
gsmendoza, line 87:
```suggestion
# Still request "devise". Solidus will skip installing it but include the seeds.
```

bin/sandbox
```diff
@@ -68,10 +66,11 @@ fi
 
 cd ./sandbox
 cat <<RUBY >> Gemfile
-gem 'solidus', github: 'solidusio/solidus', branch: '$BRANCH'
+gem 'solidus', github: 'solidusio/solidus', branch: '$SOLIDUS_BRANCH'
 gem 'rails-i18n'
 gem 'solidus_i18n'
 
+gem "solidus_auth_devise", "~> 2.5" # This extension contains a user decorator, so it must be loaded first
```
gsmendoza, line 73:
```suggestion
gem "solidus_auth_devise", "~> 2.5" # This extension contains a user decorator, so it must be loaded first.
```

gsmendoza: Also, I'm not sure if my understanding of the commit message. Please see below my updated version of the commit message, where I make the references to solidus_auth_devise more explicit. Is it correct?

> Fix bin/sandbox script generator

> Because the solidus_auth_devise extension decorates the User class it needs to
be listed in the Gemfile before the extension. Otherwise, the User class will
not be found.

> A flag will still be given to the Solidus generator so that includes
solidus_auth_devise's seeds. It will skip installing it twice however.

bin/sandbox
```diff
@@ -85,11 +84,12 @@ unbundled bundle install --gemfile Gemfile
 
 unbundled bundle exec rake db:drop db:create
 
+# Still request "devise". Solidus will skip installing it but include the seeds
 unbundled bundle exec rails generate solidus:install \
   --auto-accept \
   --user_class=Spree::User \
   --enforce_available_locales=true \
-  --with-authentication=true \
+  --authentication="devise" \
```
gsmendoza, line 92: @RyanofWoods I'm able to run `bin/sandbox` with the original argument `--with-authentication=true`. Can you confirm?

gsmendoza (commented):
@RyanofWoods Thanks for pointing me to this fix! It looks great, but I have some minor requests:

* Sorry for being a grammar nazi, but can you add periods to the comments that are sentences? :)  This may be more of a writing style preference, but for me, I'd prefer keeping it simple and using the same writing rules regardless of whether I'm writing source code comment or "normal" text. Relevant SO question: https://softwareengineering.stackexchange.com/questions/17766/what-are-your-thoughts-on-periods-full-stops-in-code-comments.
* I'm not sure I understood a commit message and proposed an edited version. Is it correct?
* It seems `bin/sandbox` can run fine now with `--with-authentication=true`. Can you test?

bin/sandbox
gsmendoza: @RyanofWoods Oh wait, I didn't notice the deprecation warning in the `bin/sandbox` output:

```
     warning  Using `solidus:install --with-authentication` is now deprecated. Please use `--authentication` instead (see --help for the full list of options).
```

So, the `--authentication="devise"` argument is actually better :)

However, I'm not sure I understand if it's related to this comment in the commit message: "A flag will still be given to the Solidus generator so that includes solidus_auth_devise's seeds. It will skip installing it twice however." Can you clarify?

elia: The latter is from solidus master, --with-authentication is the legacy one (valid up until 3.2)

elia (approved):

bin/sandbox
RyanofWoods: @gsmendoza We still provide the authentication flag so that it appends the devise extension seeds to the sandbox, which is quite useful as it makes the default admin account. The Solidus installer is smart enough to skip adding the extension to the Gemfile though:
https://github.com/solidusio/solidus/blob/efe2b3a20d0577413e4810c008ea564e637c36a3/core/lib/generators/solidus/install/app_templates/authentication/devise.rb

I am more than happy to improve the commit msg if you wish 🙂

gsmendoza: @RyanofWoods Oh wait, I understand the commit message now. You're saying that we still need to call `solidus:install` with the authentication flag even if `solidus_auth_devise` has been added to the Gemfile. I got confused because I thought it was connected to the change from the deprecated authentication flag to the new one.

RyanofWoods: Yes. But technically, it is not needed - it's just so a admin user gets created, which is very convenient.

Sorry for the confusion. It was then a good idea to update the commit msg 👍 I also moved the flag changing to another commit, because it was unrelated.

gsmendoza (approved):
@RyanofWoods Thanks for splitting the commit and improving the commit messages. They're much clearer now!

